Issue #1198, opened 2015-03-23
Add sql injection of websocket support for sqlmap


RicterZ:
If sql injection on a websocket server, it's hardly to use sqlmap.
I write a [proxy](https://github.com/RicterZ/websocket-injection) for it, but I think the best way is that add websocket support for sqlmap.

inquisb:
Pull requests are very much welcome :)
We can help with the sqlmap code if you need guidance.

RicterZ:
@inquisb 
Thanks for your reply :D 
If I want to add a third-party library([websocket-client](https://github.com/liris/websocket-client)), what should I do? Should I clone it to `thirdparty/` folder or else?

inquisb:
@RicterZ as far as I can tell, the library is packaged for for easy_install and pip so no need to add it to `thirdparty/`

RicterZ:
I found websocket-client is not processing errors completely, so I will improve error handling of websocket.
But now, for websocket, sqlmap already working:)

I don't think my code is prefect, so if somewhere I should modify the code, please tell me, thank you very much :D

RicterZ:
@inquisb 
I think I had finished it, please review, thx :)

inquisb:
@RicterZ in `lib/core/options.py` write a function that checks for the websocket library if the target URL starts with `ws:/`, make sure the check is also performed when `--dependencies` switch is provided, check the source code for example.

RicterZ:
@inquisb Got it, in fact, websocket protocol is `ws://` and `wss://` :D

RicterZ:
@inquisb fixed, please review.

RicterZ:
Now I can use sqlmap to sql injection:

    python sqlmap.py --url="ws://localhost?data=[sqli]" # sqli in get parameters

or 

    python sqlmap.py --url="ws://localhost" --data="*" # sqli in messages

But I think the best way is:

    python sqlmap.py --url="ws://localhost" # sqli in messages

Any way can implement it?

inquisb:
@RicterZ make a pull request of your changes and we will review it. Thanks.

stamparm:
Pull request 1206

PR #1206, opened 2015-03-26
Websocket support of sqlmap #1198

RicterZ:


brandonprry:
@RicterZ Do you have an example application you tested this against?

RicterZ:
@brandonprry 
Try this:

    python2.7 sqlmap.py -u "ws://103.238.227.183:8888/" --data="*"

stamparm:
@RicterZ it should be working now. Just one thing. If that server is yours (`103.238.227.183`) can you please update the `--data` to contain valid data?

RicterZ:
@stamparm Done.

stamparm:
I can't see the changes :)

I meant something like:

`python2.7 sqlmap.py -u "ws://103.238.227.183:8888/" --data="id=1"`

sqlmap now just blindly injects anything there (`--data=*`) without success in finding proper SQLi (it just recognizes MySQL from heuristics)

RicterZ:
Sorry, I misunderstood.
But the message format of websocket is "123456" rather than "id=1".
The sql injection vector is the whole message you send.

stamparm:
Ok. `123456` doesn't seem to be a valid value.

I get `{"text": "Failed to get table_schema of table_name: 123456, error: tuple index out of range"}`

Can you please state one valid value (that is working on the other side)?

RicterZ:
Any `table_name` field in the `information_schema.tables` is valid.
Try `tables`:

    select table_schema from tables where table_name='tables';

PR #1227, opened 2015-04-21
Add certutil write file technique for MSSQL

agix:
A native way to decode base64 is to use certutil binary as debug.exe doesn't work anymore on modern system and sometimes powershell is not here...

I tried to integrate it, sorry if I did it wrong and just let me know how to do it better.


stamparm:
@agix thx for your contribution. Had to slightly break it down (to be concise with the rest of the code).

PR #1244, opened 2015-05-10
Converting to python2.7 shebang

Xartrick:
Adding python2.7 instead of python.

### DIFF
--- a/extra/__init__.py
+++ b/extra/__init__.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2.7
 
 """
 Copyright (c) 2006-2015 sqlmap developers (http://sqlmap.org/)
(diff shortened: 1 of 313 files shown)


stamparm:
This won't go into official repository. sqlmap is supported by both python 2.6 and python 2.7.

stamparm:
I am wondering why you proposed this in the first place? You want accept the changes to 313 files (all *.py). There must be a really good reason.

Xartrick:
Can edit line to : `#!/usr/bin/env python2` instead.

I installed sqlmap on my Arch Linux and it cannot run using `./sqlmap.py`, it uses `#!/usr/bin/env python` who's Python 3 on my installation...

stamparm:
`python2` is not recognized everywhere as an alias to Python v2.x
screenshot - 05112015 - 09 27 12 am

PR #1323, opened 2015-08-03
Add --crawl-file option to specify output filename

ctshanghai:
Currently, there is no way to explicitly set the output file of crawling results.  For some automated tooling utilizing that feature, it is a bit more convenient to be able to control the output file name.

### DIFF
--- a/lib/core/target.py
+++ b/lib/core/target.py
@@ -508,7 +508,9 @@ def _setResultsFile():
         return
 
     if not conf.resultsFP:
-        conf.resultsFilename = os.path.join(paths.SQLMAP_OUTPUT_PATH, time.strftime(RESULTS_FILE_FORMAT).lower())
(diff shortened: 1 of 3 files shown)


stamparm:
Two things:

1) We already have tons of switches
2) sqlmap is not a crawling tool
3) If you opened a new Issue request I would put the prompt and asking where you want to store the crawling results

ctshanghai:
That's fine, I'll close this pull request.  The issue with prompting is that it's requiring user interaction, which eliminates the usefulness in the context of automation.

stamparm:
@ctshanghai can you please in future open a new Issue request, then we can discuss and afterwards if we agree you can do the pull request

ctshanghai:
@stamparm of course, no problem.

Cheers.

Issue #1341, opened 2015-08-17
Unhandled exception (#c9931c51)


sqlmapreporter:
```
sqlmap version: 1.0-dev-ce64d97
Python version: 2.7.3
Operating system: posix
Command line: sqlmap.py --pickled-options gAJjbGliLmNvcmUuZGF0YXR5cGUKQXR0cmliRGljdApxACmBcQEoVQpjcmF3bER
Technique: None
Back-end DBMS: None (identified)
```
```
Traceback (most recent call last):
  File "sqlmap.py", line 85, in main
    initOptions(cmdLineOptions)
  File "lib/core/option.py", line 2468, in initOptions
    _mergeOptions(inputOptions, overrideOptions)
  File "lib/core/option.py", line 2061, in _mergeOptions
    inputOptions = base64unpickle(inputOptions.pickledOptions)
  File "lib/core/convert.py", line 73, in base64unpickle
    retVal = pickle.loads(base64decode(bytes(value)))
  File "lib/core/convert.py", line 24, in base64decode
    return base64.b64decode(value)
  File "usr/local/lib/python2.7/base64.py", line 76, in b64decode
    raise TypeError(msg)
TypeError: Incorrect padding
```

stamparm:
It seems that somebody provided deliberatelly invalid value for `--pickled-options`. Checked everywhere else and this seems to be the most logical (at this moment) explanation for this issue

PR #1342, opened 2015-08-17
Fix for #1341

stormwin:
Added try-except to catch exception when base64 string is not valid and cannot be decoded.
Also fixed unreporeted bug. Added another try-except to catch exception when object cannot be deserialized from pickle (base64 string is valid but it is not pickle-deserializable object after base64 decode)

### DIFF
--- a/lib/core/convert.py
+++ b/lib/core/convert.py
@@ -12,6 +12,7 @@ import sys
 
 from lib.core.settings import IS_WIN
 from lib.core.settings import UNICODE_ENCODING
+from lib.core.data import logger
 
 def base64decode(value):
     """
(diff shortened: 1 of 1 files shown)


stamparm:
Not valid/complete fix. For some reason the `=` has been stripped (possibly during the sqlmapapi.py call).

```
>>> "gAJjbGliLmNvcmUuZGF0YXR5cGUKQXR0cmliRGljdApxACmBcQEoVQpjcmF3bER".decode("base64")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/encodings/base64_codec.py", line 42, in base64_decode
    output = base64.decodestring(input)
  File "/usr/lib/python2.7/base64.py", line 321, in decodestring
    return binascii.a2b_base64(s)
binascii.Error: Incorrect padding

>>> "gAJjbGliLmNvcmUuZGF0YXR5cGUKQXR0cmliRGljdApxACmBcQEoVQpjcmF3bER=".decode("base64")
'\x80\x02clib.core.datatype\nAttribDict\nq\x00)\x81q\x01(U\ncrawlD'
```

stamparm:
@stormwin or, maybe you caused that error deliberately :) ??

stormwin:
if you are asking for issue that is reported, i'm not the reporter. I just have some free time and that issue looks 'easy fix' for me... Python is not my good side, but to write 'try except' is not that complex :)

PR #1351, opened 2015-08-22
fix removeDynamicContent bug

KxCode:
double re.escape() in "findDynamicContent" function and "removeDynamicContent" function leads an bug in finding dynamic content.


stamparm:
This caused more bugs (e.g. #1367)

PR #1403, opened 2015-09-10
Add support for --rw-dirs <rwDirs> cmd line argumant / conf file option

ejtaal:
This allows one to fully automate commands where the default
upload directories cannot be used, as in:

$ sqlmap --batch --rw-dirs='/var/www/html/uploads' --os-cmd="ls;id" \
	-u 'http://nullbyte/kzMb5nVYJw/420search.php?usrtosearch=ramses'

### DIFF
--- a/lib/parse/cmdline.py
+++ b/lib/parse/cmdline.py
@@ -566,6 +566,11 @@ def cmdLineParser():
                             help="Remote absolute path of temporary files "
                                  "directory")
 
+        takeover.add_option("--rw-dirs", dest="rwDirs",
+                            help="Remote absolute path of writable "
+                                 "directories for overriding defaults "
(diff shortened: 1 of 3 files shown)


stamparm:
Don't take me wrong, but including one more option like this is uberkill.

If you have suggestions for default directories please say so. Otherwise, there is a shortcut for this:

`--batch --answer="writable=/var/www/html/uploads"`

p.s. I'll make a patch for `--answer` to also accept multiple directories this way

stamparm:
With the latest commit #c4f9e66 user can provide multiple answers like:

`--batch --answer="/var/www/html/uploads,/var/www/wordpress/images"`

ejtaal:
Clearly I didn't understand the --answer feature properly, many thanks for your updates :)

PR #1414, opened 2015-09-14
Added commands stop, kill, list to API client

daremon:
Stop & kill were straightforward. For list:
- I removed the permission check (admin_id)
- I also list the task status

I kept it backwards compatible: you can still call list with the admin taskid and get the simple list back.

### DIFF
--- a/lib/utils/api.py
+++ b/lib/utils/api.py
@@ -698,12 +698,12 @@ def client(host=RESTAPI_SERVER_HOST, port=RESTAPI_SERVER_PORT):
 
     while True:
         try:
-            command = raw_input("api%s> " % (" (%s)" % taskid if taskid else "")).strip()
+            command = raw_input("api%s> " % (" (%s)" % taskid if taskid else "")).strip().lower()
         except (EOFError, KeyboardInterrupt):
             print
             break
 
-        if command.lower() in ("data", "log", "status", "stop", "kill"):
+        if command in ("data", "log", "status", "stop", "kill"):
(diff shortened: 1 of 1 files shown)


stamparm:
Everything looks ok other than the removal of `is_admin` check. That is a security check so the end user couldn't mess up with other tasks (list, kill,...) if he doesn't know the magic "nonce" `Admin ID`.

daremon:
OK, do you want me to put the check back and require the `admin_id` when executing `list`?

It would be nice if the server could tell you your tasks, but I can't think of a simple way to do it.

stamparm:
Give me a little time. Will push the `simple way`

stamparm:
With the latest commit you'll be able to do the `admin` stuff without knowing the `ADMIN id`. Only (safety) catch is that the tasks the (non-admin) user will be able to list/flush will be those that have the same `remote_addr` (same origin as the user itself).

One thing. In case that the user is hooked up for the first time, to avoid the `new` for `list` you can provide whatever you want for the task_id (e.g. `0`). I can do this afterwards in case that I am not clear enough

daremon:
I think it works as expected now. BTW I changed the list command output to include the status of each task (line 386).

lordpato:
Could you please check this? :

  File "api.py", line 387
    for x in list(DataStore.tasks)}
      ^
SyntaxError: invalid syntax

stamparm:
That is not there. Please clone from the start.

Bye

PR #1449, opened 2015-10-06
Create hexencode.py

DrWhatNoName:
Useful for applications which mite use hex encoding on data before sent to the server.


stamparm:
Tamper scripts are for bypassing WAFs/IPS/IDS. This encoding stuff is usually being done in the `--eval`.

PR #1543, opened 2015-11-18
enhance recognition of "baiduyun" in waf/baidu.py

Wyc0:
The former recognition is not that accurate because "Yunjiasu Web Application Firewall (Baidu)" didn't have a specific trait back then.
But now it has the "server: yunjiasu-nginx" headers which makes it more easily to recongnize.
And I tested serval sites using this WAF, the old baidu.py doesn't work besides with "__cfduid" in cookie it was alwasy mistoken for Cloudflare,by adding new new regex pattern in "waf/baidu.py" the problem is now solved.


stamparm:
@Wyc0 thx :)

Wyc0:
love sqlmapProject much, thx XD

PR #1565, opened 2015-11-26
fix detect bug

jjf012:
for example
if result='11111111111.abcd'
current can't change


stamparm:
If you are sure that this is fixing a real case, please attach a complete
error message related or contact me at miroslav@sqlmap.org

PR #1678, opened 2016-01-22
add --servername option to support various of  bottle server adapter

dozysun:
add --servername option to support various of  bottle server adapter,   improve server concurrent


stamparm:
Will take a look tomorrow

stamparm:
shouldn't this option be called `--adapter`/`adapter`?

dozysun:
yes,  it should be,   need i to change it or you do it ?

stamparm:
I can do it later today (if you don't make it)

dozysun:
have changed it already

PR #1681, opened 2016-01-26
fix logical error

imoldman:
Commit c7ef9429ae80d31bfe79b5ed9d49d08b77530049 make boolean-based blind test invalid, fix it.

### DIFF
--- a/lib/controller/checks.py
+++ b/lib/controller/checks.py
@@ -793,7 +793,7 @@ def checkFalsePositives(injection):
                 retVal = None
                 break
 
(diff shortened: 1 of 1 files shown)


stamparm:
This is wrong. Invalid comparison like `AND foo bar` should not return `True` ever.

You are now fine-tuning this check for your case, while you are crippling the sqlmap for all other cases. Please contact me privately

stamparm:
p.s. I am pretty sure that you are having problem with `OR boolean` but your patch is not the solution

PR #1700, opened 2016-02-08
add support for parsing Excel (xlsx) output

arthepsy:
This adds support for SQLi testing in cases when webpage outputs Excel file (application/vnd.ms-excel). It works by converting all worksheets to HTML.

Have been using this successfully for a week with several different outputs. 

Currently, it depends on installed `openpyxl` library. This is not perfect and requires either better error message or inclusion in `thirdpary` directory. Feedback?

### DIFF
--- a/lib/request/basic.py
+++ b/lib/request/basic.py
@@ -5,6 +5,7 @@ Copyright (c) 2006-2016 sqlmap developers (http://sqlmap.org/)
 See the file 'doc/COPYING' for copying permission
 """
 
+import cgi
 import codecs
 import gzip
 import logging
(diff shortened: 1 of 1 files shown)


stamparm:
Well, this is really a special case. Introducing new 3rd party dependency for such special case...

arthepsy:
Is there a better way without 3rd party dependency? Currently, without this, sqlmap just considers Excel-outputting webpages as "unstable" and no injection is found.

In any case, I was tending to "better error message" option, but I am not sure what would be the correct way to implement it. Raising exception seems a bit wrong. Maybe just log error and pass.

If you are interested in this feature, give me some pointers and I'll polish it.

stamparm:
You've done a great job for your case. Also, this should be a (google searchable) reminder for people that will have such case. Other than that - I am not sure that this should go into the master (otherwise, why not implement the pdf parsing, rar uncompression, certificate parsing, etc.)

arthepsy:
Yes, while I was looking over the code, I was thinking exactly that - why there isn't any other parsing implemented. I get your point about core/master. Actually, I would love to have this implemented as a plugin, but there isn't framework (or I didn't find) for output parsing before analysis. Maybe I can suggest such feature? Then keeping a plugin separate wouldn't be an issue at all.

PR #1762, opened 2016-03-14
Add 'randomcommentsextended' tamper script.

squarefractal:
This is similar to the randomcomments tamper script, but adds random strings within the comments.

### DIFF
--- a/tamper/randomcommentsextended.py
+++ b/tamper/randomcommentsextended.py
@@ -1,0 +1,49 @@
+#!/usr/bin/env python
+
+"""
+Copyright (c) 2006-2016 sqlmap developers (http://sqlmap.org/)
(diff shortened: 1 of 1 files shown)


stamparm:
`randomcomments` should be removed because it is based on false assumption that /**/ can be put inside the SQL keyword. Can you please give one example where yours or `randomcomments` can work (which DBMS)?

stamparm:
Only case where this could work if the WAF removes the `/*...*/` before sending the whole query to the DBMS. But I really doubt that this is realistic case.

squarefractal:
I'm contributing this indeed keeping a broken WAF in mind, which is doing that exact thing you mention.

stamparm:
No hard feelings :)

PR #1776, opened 2016-03-23
Use LIMIT %d OFFSET %d insetad of LIMIT %d,%d in blind sql injection on MySQL dbms

globisz:
Well, sometimes webapp/waf filter comma characters, and despite usage of commalessmid tamper, table enumeration becomes impossible. But because MySQL supports LIMIT %d OFFSET %d clause its can be "bypassed".

It works in the case when you provide sqlmap name of database, table and column(s) that you want to enumerate. Because sqlmap checking access to information_schema first (or i missed something and this webapp also filtered "information_schema" string), using query with comma characters, this patch will not give you the possibility to enumerate table just by giving its name.
However, this patch worked in my case well, and should not break anything.

### DIFF
--- a/lib/core/agent.py
+++ b/lib/core/agent.py
@@ -886,7 +886,11 @@ class Agent(object):
         fromFrom = limitedQuery[fromIndex + 1:]
         orderBy = None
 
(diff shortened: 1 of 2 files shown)


stamparm:
This should go into the tamper script

stamparm:
With the latest revision/commit: `--tamper=commalesslimit`

PR #1805, opened 2016-04-09
Add initial .editorconfig file

walterdolce:
Given [it's mentioned in the FAQ page](https://github.com/sqlmapproject/sqlmap/wiki/FAQ#can-i-contribute-occasionally-to-the-development):

> Avoid tabbing, use four blank spaces instead.

And that it's not currently being used within the project, I thought to add an initial .editorconfig file defining just the spacing policy for `.py` files. This should let (EditorConfig compliant) IDEs used by contributors do the right thing when it comes to spacing.

I didn't add anything else as this is my first commit and look at the project as a whole, so I'd leave this to the core project maintainers ;)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,3 +4,4 @@ root = true
 
 [*.py]
 indent_style = space
+indent_size = 4
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,6 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+root = true
+
+[*.py]
+indent_style = space


stamparm:
Well, we don't use EditorConfig. Not sure what to say else

walterdolce:
It is just a file with instructions for your IDE to follow while you work on the project. 
In this case, it will make sure to convert tabs to [4] spaces when working on `*.py` files.

stamparm:
I am aware of its functionality. Though, "core developers" here don't use this. Can't merge something that is not meant to be part of the project.

PR #1843, opened 2016-05-05
Change IP for new request

Belladona-c0re:
Bypass Firewall protections that ban IP when you submit some sql injection attempt

### DIFF
--- a/tamper/bypassip.py
+++ b/tamper/bypassip.py
@@ -1,0 +1,29 @@
+#!/usr/bin/env python
+
+"""
+Copyright (c) 2006-2015 sqlmap developers (http://sqlmap.org/)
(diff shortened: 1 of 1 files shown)


stamparm:
Several things are problematic here:

1) using `pkill` like this SIGHUPs non-tor processes too (e.g. `indicator-services-start`, `gvfs-udisks2-volume-monitor`...)
2) tamper scripts are for modification of payloads, not for anything else (things like `--eval` exists which can be used for something like you are trying to achieve)
3) unprivileged users (i.e. non-root) can't do this, hence, sqlmap would just drop dead here

p.s. closing this down with a way around:

```
... --eval "import os; os.system('pkill -sighup tor')"
```

PR #2089, opened 2016-08-09
make sqlmap use baidu search engine

5a43:
currently, sqlmap only support google engine, but Google is blocked by China GFW. In China, maybe use baidu is more easier than google. So i created this pull request.


stamparm:
There are couple of problem here:

1) sqlmap failbacks to Duckduckgo if Google fails, so the statement "sqlmap only support google engine" is not true
2) You have some problematic lines in your pull request (e.g. "Because sqlmap only support GoogleDork. I fork original project and add BaiduDork support." from https://github.com/sqlmapproject/sqlmap/pull/2089/commits/ab7e9cc289e2aba41d4519caa736b92e884e6f32) which can't go to the master
3) Switches and/or options can't be introduced without a specific request. There are already tons of switches/options
4) Why not incorporating Baidu the same way as for the Duckduckgo? When Google fails, ask user whether he wants to use Duckduckgo or Baidu as other methods

PR #2169, opened 2016-09-14
baidu search engine support

5a43:
add baidu(www.baidu.com) search engine support.  

In china, google can't be accessed directly if you don't have a proxy or VPN. Original code will make sqlmap to exit if google can't be accessed or connected. I changed the action, and make the code to use google search only when google is reachable.   The added code of baidu search support will do an extra "finding real url". Because in result of baidu, we can't see the real url, we have to access the result one by one to get the real URL. At last, I removed similar/duplicate URLs from retVal because baidu's results often include some similar or duplicate ones.


omg2hei:
cool~

PR #2350, opened 2017-01-16
Fixes OSError in API

qnrq:
Fixes OSError raised by sqlmapapi.py when started as a server outside the script working directory and without sqlmap in a global path:

  File "/usr/local/sqlmap/lib/utils/api.py", line 167, in engine_start
    self.process = Popen(["sqlmap", "--pickled-options", base64pickle(self.options)], shell=False, close_fds=not IS_WIN)
  File "/usr/lib/python2.7/subprocess.py", line 710, in __init__
    errread, errwrite)
  File "/usr/lib/python2.7/subprocess.py", line 1335, in _execute_child
    raise child_exception
OSError: [Errno 2] No such file or directory

### DIFF
--- a/lib/utils/api.py
+++ b/lib/utils/api.py
@@ -140,6 +140,7 @@ class Task(object):
         self.options.api = True
         self.options.taskid = taskid
         self.options.database = Database.filepath
+        self.cwd = os.getcwd()
 
         # Enforce batch mode and disable coloring and ETA
         self.options.batch = True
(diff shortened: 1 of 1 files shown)


stamparm:
Can you give one meaningful example when this is the case? API expects to be inside the same directory as sqlmap. It shares all the libraries in the first place.

This all looks like your environment problem

stamparm:
After a second look I believe I know what you wanted to do. Now commuting, out of real keyboard

qnrq:
> Can you give one meaningful example when this is the case? API expects to be inside the same directory as sqlmap. It shares all the libraries in the first place.
>
> This all looks like your environment problem

For example:

```bash
cd / && python /tmp/sqlmap/sqlmapapi.py --server
```

It could be solved with the env but it's not always practical.

PR #2369, opened 2017-01-23
Added payload for forced comments on boolean-based blind injections

Ark444:
Added payload for forced comments on boolean-based blind injections for WHERE or HAVING clause.


stamparm:
This is already covered (https://github.com/sqlmapproject/sqlmap/blob/master/xml/payloads/boolean_blind.xml#L257-L295). Level has to be raised to >=3 for it to be used

PR #2374, opened 2017-01-31
New space 2 more comment bypass

anarcoder:
[add] new space 2 more comment bypass


stamparm:
Adding nothing new than the regular `/**/` (e.g. `space2comment` or `modsecurityversioned`)

anarcoder:
I make this modification because "/**/" is blocked some times.. and using /**_**/ is old school trick to bypass this, usually is used in buffer overflow payloads... but no problems =))

PR #2389, opened 2017-02-07
REST API: command line processing on the server side

kojenov:
We are writing a sqlmap API client for verifying SQLi exploits. The exploits come to us in the form of sqlmap command line options. For example:
`-u 'http://server/query?sort=tag' -p sort --dbms=mysql --level=5 --technique=T --time-sec=1 -b --answers="redirect to=N,extending provided risk=N,keep testing the others=N"`

Currently, the API's _start_ function expects a JSON. This is trivial to achieve for the sample API client (sqlmapapi.py -c) since it shares the codebase with sqlmap.py and has access to _cmdLineParser()_, but for our application, it would be very non-trivial.

I added support for parsing the command line options on the API server side, so the API client can stay completely ignorant of the internal sqlmap option representation and processing.

### DIFF
--- a/lib/parse/cmdline.py
+++ b/lib/parse/cmdline.py
@@ -38,7 +38,7 @@ from lib.core.shell import clearHistory
 from lib.core.shell import loadHistory
 from lib.core.shell import saveHistory
 
-def cmdLineParser(argv=None):
+def cmdLineParser(argv=None, interactive=True):
     """
     This function parses the command line parameters and arguments
     """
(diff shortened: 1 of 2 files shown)


lib/parse/cmdline.py
```diff
@@ -973,7 +973,7 @@ def _(self, *args):
 
     except SystemExit:
         # Protection against Windows dummy double clicking
-        if IS_WIN:
+        if IS_WIN and interactive:
```
kojenov: API server on windows would "hang" waiting for user to push a button

kojenov:
Our use case is simply to determine whether the supplied sqlmap attack is successful or not, and goes like this:
1. Start the scan
2. Wait until the scan is complete - /scan/<taskid>/status
3. Check the result - /scan/<taskid>/data. Here, we don't do a lot of parsing; we just check if data is non-empty, which indicates the attack succeeded.

So overall, the tool is pretty lightweight, and tries to rely on sqlmap code as much as possible. That's why we'd prefer sqlmap to parse the initial command line in step 1.

PR #2417, opened 2017-02-27
Fixed some grammar in a few files

Ekultek:
Fixed some grammatical flaws in a few files, added `.idea/` to `.gitignore` because pycharm creates a .idea directory that nobody likes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,4 @@ output/
 .sqlmap_history
 traffic.txt
 *~
-idea/
+.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ output/
 .sqlmap_history
 traffic.txt
 *~
+idea/
--- a/lib/core/common.py
+++ b/lib/core/common.py
@@ -591,10 +591,10 @@ def paramToDict(place, parameters=None):
                       or re.search(r'\A9{3,}', _) or re.search(r'\A-\d+\Z', _) or re.search(DUMMY_USER_INJECTION, _))\
                       and not parameter.upper().startswith(GOOGLE_ANALYTICS_COOKIE_PREFIX):
                         warnMsg = "it appears that you have provided tainted parameter values "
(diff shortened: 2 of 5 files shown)


stamparm:
There are lots of problems with your "grammar fixes". I am not sure where to begin:

1) There is nothing wrong with the "Please, always use only valid parameter values so sqlmap could be able to run properly" because ""Could" is used to express possibility or past ability as well as to make suggestions and requests". You are proposing "...so sqlmap will be able to run properly" which doesn't have any sense at all.
2) I don't see a grammar problem with "most probably web server instance hasn't recovered yet" while you are proposing "most probable cause is that, the web server" <- why is here a ','? Also, usage of `that` in this context is redundant
3) "most probably caused by inability" is ok. Your "most probable cause is by inability" is totally wrong
...

Ekultek:
Well for one "please always use only valid parameters so sqlmap could be able to run properly" doesn't make any damn sense, it should be "please use only valid parameters for sqlmap to run successfully" that's actual English.

Two "most probably" isn't correct English, you can Google an English text book and look. "Most probable", is the correct way to say "most probably"

Three "most probable cause, by inability" makes sense, again English, I know it's not your first language but it's not that difficult to run your sentences through MS Word or equivalent.

Ekultek:
Four errornous is spelt erroneous.

Five "please wait few minutes" is incorrect grammar, should be "please wait a few minutes"

I mean dude if you want me to call you out on your grammar I will. But don't sit there and tell me I'm wrong when I'm not.

stamparm:
I've changed the "few minutes" to "a few minutes" in https://github.com/sqlmapproject/sqlmap/commit/7ea524800ad0d1df172896a19b85d25dfd7a2945

stamparm:
With latest commit changed from "most probably" to "most likely"

stamparm:
I am not going to take any more suggestions from here into the consideration. Thank you

PR #2418, opened 2017-02-28
IP address regex updated to not provide a false positive

Ekultek:
Updated the IP address regex from `\b\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}\b` to `^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])$` the latter will not provide a false positive. The old style regex would match anything as long as it's in the format `0.0.0.0` including `999.999.999.999` see [here](https://regex101.com/r/hcQKT9/5) for examples. The newer style regex will not match anything over `255.255.255.255` see [here](https://regex101.com/r/hcQKT9/3) for examples of the new regex.

Basic examples:

```
>>> import re
>>> IP_REGEXS = [r"\b\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}\b", r"^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])$"]
>>> IP_ADDRS = ["127.0.0.1", "255.255.255.255", "999.999.999.999", "257.12.34.543", "12.09.87.24"]  # Only the first two should match (127.0.0.1, 255.255.255.255)
>>> for item in IP_REGEXS:
	for ip in IP_ADDRS:
		if re.match(item, ip):
			print("{} matched {}".format(item, ip))
		else:
			print("{} did not match {}".format(item, ip))

			
\b\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}\b matched 127.0.0.1
\b\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}\b matched 255.255.255.255
\b\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}\b matched 999.999.999.999
\b\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}\b matched 257.12.34.543
\b\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}\b matched 12.09.87.24
^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])$ matched 127.0.0.1
^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])$ matched 255.255.255.255
^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])$ did not match 999.999.999.999
^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])$ did not match 257.12.34.543
^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])$ did not match 12.09.87.24  # Did not match because ..09.. is not a valid IP address, valid would just be 9
>>> 
```

### DIFF
--- a/lib/core/settings.py
+++ b/lib/core/settings.py
@@ -109,7 +109,7 @@ DUMMY_SEARCH_USER_AGENT = "Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:49.0) Geck
 TEXT_TAG_REGEX = r"(?si)<(abbr|acronym|b|blockquote|br|center|cite|code|dt|em|font|h\d|i|li|p|pre|q|strong|sub|sup|td|th|title|tt|u)(?!\w).*?>(?P<result>[^<]+)"
 
 # Regular expression used for recognition of IP addresses
-IP_ADDRESS_REGEX = r"\b\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}\b"
+IP_ADDRESS_REGEX = r"^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])$"
 
 # Regular expression used for recognition of generic "your ip has been blocked" messages
 BLOCKED_IP_REGEX = r"(?i)(\A|\b)ip\b.*\b(banned|blocked|block list|firewall)"


stamparm:
I've deliberately didn't use that specific regex because I was pretty sure that it won't cause any "problems" (that same regex is used only in brute forcing of web page root path - expecting error messages)

Please find me a single page where sqlmap has problems and I'll immediately merge the request

stamparm:
p.s. making a boundary around the regex (as you provided) expects that IP address is the single element inside the line (or whole content). As you've correctly noticed that English is not my "mother's" language I can notice that you are not a developer, right?

Ekultek:
That makes sense actually, I was just trying to not provide a false positive for it. If for some reason the IP is lets say honey potted, your regex will match anything. But it makes sense being just for bruteforcing. If I do find a specific issue with it, I'll open another pull request thanks

stamparm:
Merged and "corrected"

Ekultek:
Why boundary though? The way the IP would be provided would be in a whole as a JSON info, dict, or list wouldn't it?

stamparm:
Boundaries are used to find "boundaries" between words. If there is an IP address inside a content there should be boundaries (e.g. `foo1.2.3.4aaaa` contains IP address without boundaries, while `foo 1.2.3.4 aaaa` contains IP address with boundaries). I am interested only in word-alike recognitions (has to contain boundaries for correct match)

stamparm:
p.s. just to prevent further question(s), sole `1.2.3.4` inside a line is considered as "bounded" too. The same thing is with e.g. `>1.2.3.4<` or `"1.2.3.4"`. They all contain boundaries. Please google for regex boundaries

PR #2481, opened 2017-04-14
Fix misuse of flags in re.sub() calls

jwilk:
The 4th argument of `re.sub()` is maximum number of substitutions, not flags.

Found using [pydiatra](https://github.com/jwilk/pydiatra).


stamparm:
@jwilk thx for this. Subtle bug :). I pushed the more "appropriate" version (to be consistent with the rest of the code)

PR #2506, opened 2017-04-27
add more keywords

carterjones:
These keywords are used in various tests by sqlmap, so adding them to this list
helps when running tamper scripts that rely on the accuracy of keyword lists.


stamparm:
Those new items are (MySQL) function names (e.g. `EXP`) and identifiers (e.g. `INFORMATION_SCHEMA`) not keywords. Keywords are incorporated into the SQL parser and those MUST not be mutable.

PR #2616, opened 2017-07-18
Implementation of a custom injection char

vdbaan:
In normal circumstances the asterisks is used as injection character. However in those cases where this character is also used at different places (for example in the cookie), this allows you to use a different character to identify the injection point.
e.g.
sqlmap.py --flush-session --batch -u <URL>/injection.php?p=*
could become:
sqlmap.py --flush-session --batch --custom-injection-char='@' -u <URL>/injection.php?p=@

### DIFF
--- a/lib/core/agent.py
+++ b/lib/core/agent.py
@@ -101,7 +101,7 @@ class Agent(object):
         if place == PLACE.URI or BOUNDED_INJECTION_MARKER in origValue:
             paramString = origValue
             if place == PLACE.URI:
-                origValue = origValue.split(CUSTOM_INJECTION_MARK_CHAR)[0]
+                origValue = origValue.split(conf.customInjectionChar)[0]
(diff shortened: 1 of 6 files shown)


stamparm:
There is absolutely no need for this. Havij-like marker %INJECT_HERE% or %INJECT HERE% is also supported (Reference: https://github.com/sqlmapproject/sqlmap/wiki/Usage#uri-injection-point)

vdbaan:
I just tried the marker on the following query:
sqlmap --cookie="phpID=whatever*123" -u "http://example.com/url/%INJECT_HERE%"
but the astrisks in the cookie disappears, which shouldn't happen, but does as it's the default injection marker.

stamparm:
Will patch later today (missing asterisk)

vdbaan:
thank you

PR #2618, opened 2017-07-19
Mark steps in HAR file

lphuberdeau:
I've included some extended attributes to the entity blocks to allow tracking which request is associated to which test and better follow through the flow.

I also fixed a few minor issues related to the document format that caused errors while loading.

I also added handling for "HTTP redirect", which was misparsed (unhandled) and reported as a 200 instead.


stamparm:
Now I get invalid syntax check at http://www.softwareishard.com/har/viewer/:

PR #2663, opened 2017-08-20
`response code=404` handle modify and `https with -r` handle modify

3xp10it:
detail:
<a href ="http://3xp10it.cc/web/2017/08/20/sqlmap%E5%AF%B9%E7%8A%B6%E6%80%81%E7%A0%81404%E5%A4%84%E7%90%86%E7%9A%84bug/">sqlmap 404 handle bug</a>

When sqlmap got 404 code from the server,sqlmap will raise a exit exception and will not scan the url's sqli anymore,but some 404 code may not be a real 404 code,`dvwa` is an example,and some waf may return 404 http response code,but the body content in the page of the url is not `404` type content,so I add a handle to recgnize whether the code 404 is a real 404 or not,if the 404 is not a real 404,sqlmap will not raise the exception and go on with latter tests.

### DIFF
--- a/lib/core/option.py
+++ b/lib/core/option.py
@@ -331,7 +331,12 @@ def _feedTargetsDict(reqFile, addedTargetUrls):
                 elif not scheme and port == "443":
                     scheme = "https"
 
-                if conf.forceSSL:
+                hostPattern=host.replace(".","\.")
+                if (conf.forceSSL or re.search(r"host:[^\n\r]*%s:%s[\s\S]*referer:[^\n\r]*https://%s:%s.*" %
+                                              (hostPattern, port, hostPattern, port), request, re.I)
+                or (port != 80
+                    and re.search(r"host:[^\n\r]*%s:%s[\s\S]*referer:[^\n\r]*https://.*cdn.*" %
+                                  (hostPattern, port), request, re.I))):
                     scheme = "https"
                     port = port or "443"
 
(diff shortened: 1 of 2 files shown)


3xp10it:
another bug on https type sqli scan:
detail:
http://3xp10it.cc/web/2017/08/23/sqlmap%E4%BB%A5%E5%8A%A0%E8%BD%BD%E8%AF%B7%E6%B1%82%E6%96%87%E4%BB%B6%E6%96%B9%E5%BC%8F%E6%A3%80%E6%B5%8Bhttps%E6%9C%8D%E5%8A%A1%E5%99%A8%E6%97%B6%E5%AD%98%E5%9C%A8%E7%9A%84bug/

stamparm:
Your issue request doesn't make any sense and the code quality is really poor. When you start sqlmap against the page that issues 404 code, you are being asked whether you want to proceed. From your comments I couldn't understand where sqlmap fails in the whole workflow to suite your needs.

Also, it doesn't make any sense to distinguish chinese 404 from the rest of the world 404 (as could be seen from your pull request). Closing this down as invalid

stamparm:
With the latest commit there is a new option (`--ignore-code`):

```
$ python sqlmap.py -hh | grep "error code"
    --ignore-code=IG..  Ignore HTTP error code (e.g. 401)
```

Just use `--ignore-code=404` in your kind of cases. As said, poor quality code won't go into the repository, especially in cases like https://github.com/sqlmapproject/sqlmap/pull/2663/commits/20daabdfc808e4bb5103be98e7ad975142772dca

PR #2667, opened 2017-08-23
add strip for path variable

lzskyline:
//sry for typo

An error occurred when I used the feature "custom common column existence check".
I found it's caused by the double quotation marks which was added by "Copy as path"(a feature in Windows) automatically.
(e.g: "D:\dist\dict1.txt")
I think it's not a good idea that let user delete the double quotes manually.


stamparm:
Duplicate of #2666

PR #2690, opened 2017-09-07
Fix regex for tamper script greatest

tint0:
The original regex in greatest.py didn't work in a blind injection case that filters '>'. Payload sample from sqlmap: admin' AND ORD(MID((SELECT DISTINCT(IFNULL(CAST(schema_name AS CHAR),0x20)) FROM INFORMATION_SCHEMA.SCHEMATA LIMIT 0,1),1,1))>64 AND 'XJnE'='XJnE

Also in the formula "tamper('1 AND A > B')", B in most cases is just a number, so rewrite the request using LEAST like LEAST(A,B+1)=B+1 makes it much shorter and easier to read than GREATEST(A,B+1)=A, hence the new least.py

### DIFF
--- a/tamper/least.py
+++ b/tamper/least.py
@@ -16,21 +16,19 @@ def dependencies():
 
 def tamper(payload, **kwargs):
     """
-    Replaces greater than operator ('>') with 'GREATEST' counterpart
+    Replaces greater than operator ('>') with 'LEAST' counterpart
 
     Tested against:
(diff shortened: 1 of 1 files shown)


stamparm:
Grateful for your report and pull request. Though, I had to rewrite regexes :) (to feed my OCD midget)

tint0:
That's okay, glad to know it's fixed :)

Issue #2604, opened 2017-07-06
JSON values "true, false, null" ignored


drwetter:
Hello,

it's me again. :)

If I have a single JSON object in the POST body , keys having a value without a value in quotes -- in my cases ``true``, ``false``, ``null`` are not being tested.

```
{
   "myval": "1"
   "groupByAAAA": true
}
```

First one is being tested, second one and others as described not.
 
Thx, Dirk

stamparm:
`true` and `false` are ignored to save scanning time as it is assumed that (e.g.) JSON parameters having boolean values should not accept string values (SQLi payloads). I'll patch the workflow for `null`, though, we could further discuss for (e.g.) boolean values.

drwetter:
Hmm, _should not_ is different to _does not._ From my point as a tester I rather would like to make sure that there's no injection vector hidden. I understand the scanning time argument though thus a cmd line switch or a q asking the user (as for JSON in general) could come in handy

drwetter:
👍

stamparm:
@drwetter please take a look at #2692

PR #2692, opened 2017-09-10
Fixed json data non-string values passed to a target as strings

ikhvostenkov:
When you send JSON data with POST request consisting numeric or boolean values in original JSON body, all these values are transformed to string representations and send out to a target as strings. All requests with transformed JSON data are rejected by target server due to JSON parse errors. I've changed replacing of original values with strings alternatives for boolean and numeric values. Probably, I do not know the original idea of that logic and some additional changes are required, but right now all requests with JSON body consisting numeric and boolean values fail and tests are not possible, with that fix requests are sent correctly and tests work, please, review.

### DIFF
--- a/lib/core/target.py
+++ b/lib/core/target.py
@@ -142,8 +142,8 @@ def _setRequestParams():
                     conf.data = getattr(conf.data, UNENCODED_ORIGINAL_VALUE, conf.data)
                     conf.data = conf.data.replace(kb.customInjectionMark, ASTERISK_MARKER)
                     conf.data = re.sub(r'("(?P<name>[^"]+)"\s*:\s*"[^"]*)"', functools.partial(process, repl=r'\g<1>%s"' % kb.customInjectionMark), conf.data)
-                    conf.data = re.sub(r'("(?P<name>[^"]+)"\s*:\s*)(-?\d[\d\.]*)\b', functools.partial(process, repl=r'\g<1>"\g<3>%s"' % kb.customInjectionMark), conf.data)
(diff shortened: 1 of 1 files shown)


stamparm:
Caused by #2604 //cc: drwetter

stamparm:
@ikhvostenkov basically, you say that server rejects non-integer data on parameters having integer values. I don't see how is this sqlmap's problem. You want to test the SQLi on such parameters without changing the (e.g.) integer value to string value (to carry SQLi payload in the first place).

With this push you are sending inherently invalid format (e.g. `"p": 1 OR 1=1`) which from my point of view could in best place be interpreted at the server side as integer value `1`.

If you are seeing target invalidating SQL injected parameter values "numeric or boolean" that's actually excellent because you know that any form of SQLi attempts on those will fail

ikhvostenkov:
@stamparm Ok, I understand the logic and that makes sense, but the problem is before start tests, you check the provided URL with provided data. This data will be modified by sqlmap and test call will fail with 4xx and tests will not start at all. So then probably we should not fail test connection if 4xx returned or add additional logic in the test connection to do not modify initial JSON data. What do you think on that?

stamparm:
That makes sense. I'll fix that.

ikhvostenkov:
@stamparm Thanks a lot!

bradencross:
I am having an issue related to this same problem. If I understand correctly sqlmap tries to replace a boolean or number value with a string for JSON data. However I find that it does this incorrectly, for example if I have the following option: --data="{"\isBoolean": true}". Sqlmap will modify the isBoolean parameter in a way like: {"isBoolean": \ "true AND 5506=5201\ "} (spaces after the \ is not there in actuality) and what it should be sending is {"isBoolean": "true AND 5506=5201"}. The \ should not be in the sent POST request.

PR #2732, opened 2017-10-05
Minor optimisation

syedafzal:
Hi, 
Minor optimisation, moving len(payload) outside the while loop. Reducing number of function calls per iteration.  
Please revert if any concerns.


stamparm:
There is nothing to be optimized here. Can't consider this as an optimization in this non-resource consuming piece of code

PR #2733, opened 2017-10-05
Add boolean-blind for postgreql in stacked queries

NetSpida:
The patch is based on time-based blind exploitation, but can provide much faster results.


stamparm:
Already inside (https://github.com/sqlmapproject/sqlmap/blob/master/xml/payloads/boolean_blind.xml#L1366-L1386). This is nothing new

PR #2791, opened 2017-11-21
Add new tamper script which can replaces instances like 'IFNULL(A, B)...

VitalySalnikov:
Hello, I created very similar tamper script like already exists here https://github.com/sqlmapproject/sqlmap/blob/master/tamper/ifnull2ifisnull.py

In my case WAF blocked my payloads if there are exists IFNULL() and IF() functions, so I just little modified **ifnull2ifisnull** tamper script.

> tamper('IFNULL(1, 2)')
'CASE WHEN ISNULL(1) THEN (2) ELSE (1) END'

I guess it could be useful for other too.

### DIFF
--- a/tamper/ifnull2casewhenisnull.py
+++ b/tamper/ifnull2casewhenisnull.py
@@ -1,0 +1,65 @@
+#!/usr/bin/env python
+
+"""
+Copyright (c) 2006-2017 sqlmap developers (http://sqlmap.org/)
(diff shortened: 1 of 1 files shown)


stamparm:
I'll merge this one, but compared to original I don't see anything new. As a matter of fact, `CASE WHEN` is filtered / blacklisted more than `IFNULL` from my experience

PR #2807, opened 2017-11-27
Added alpine-based Dockerfile + documentation

khast3x:


stamparm:
Sorry, this won't go here as we don't endorse similar "docker-alike" setups. This doesn't prevent you to share this on your Github pages

PR #2817, opened 2017-12-05
Batch file to run on windows commandline

ZishanAdThandar:
Batch file to runnable on command prompt with **sqlmap** command after setting up Windows environment variable path to sqlmap and python27 directory.


stamparm:
I can't accept something like this to the trunk. Please use it at your side

PR #2903, opened 2018-02-01
Feature/compatible with py3/print

akun:
* Compatible with Python 3 step by step except `thirdparty`.
* run: `2to3 . -f print -n -w`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 *.py[cod]
+*.sw[op]
 output/
 .sqlmap_history
 traffic.txt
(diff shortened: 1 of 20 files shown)


akun:
Compatible with Python 3 step by step except `thirdparty`.

akun:
```
2to3 . -f print -n -w
```

stamparm:
I closed down other related pull requests because of two reasons:

1) When you are making this kind of pull requests (chained) it is "acceptable" to make a single (squashed) pull request than to make N different ones
2) There is no partial Python 3 coverage. Either it can be run in Python 3 or it can't be run. I am not sure that these couple of 2to3 done the job

akun:
OK. I'll make a single PR and add some test.

Compatible with Python 3 may continued for a while.

akun:
see: #2967

PR #2931, opened 2018-02-13
Create bypass360waf

tennc:
maybe bypass 360waf


ganziqim:
why version number 99999 can bypass 360waf?
it is ineffective on my computer

stamparm:
@ganziqim thank you for review. @tennc any comments?

PR #2933, opened 2018-02-14
Use system magic python library where available (e.g. Debian/Ubuntu)

LocutusOfBorg:


stamparm:
Just tried on regular Ubuntu 16.04.3.

System `python-magic`doesn't have functions that are already being utilized by sqlmap, like `magic.from_file(...)` or `magic.from_buffer(...)`. Also, no need for changes of this kind (from `thirdparty` to system ones) - as in some cases, `thirdparty` libraries were actually being fixed inside the sqlmap itself

LocutusOfBorg:
@stamparm python-magic exists only in ubuntu development release and debian testing, not available on xenial
https://launchpad.net/ubuntu/+source/python-magic
(maybe you are referring to something old and different?)

stamparm:
@LocutusOfBorg I believe that you've mixed something up. `python-magic` is alive and well in xenial (https://packages.ubuntu.com/xenial/python/python-magic), though incompatible with current sqlmap requirements. There is really no need to update anything inside `thirdparty`, except if coming from us.

Screenshots:

LocutusOfBorg:
Sorry, I should have been more verbose, sorry for that.

In artful, xenial and before, (so, up to the current stable Ubuntu and Debian release, python-magic was provided by src:file.

Now, since Buster, and Bionic (aka next Debian stable, and next Ubuntu LTS, out in some months), python-magic from src:file has been dropped, and now the python-magic is provided by the pypi module
https://packages.qa.debian.org/p/python-magic.html
https://launchpad.net/ubuntu/+source/python-magic
https://github.com/ahupp/python-magic/

there is even a compat module patch, that makes it compatible with the xenial version, but compatible too with the features requested by sqlmap.

At the end, what about trying to import the API you need and fallback to the thirdparty version if you get an AttributeError?


BTW almost all linux distro have switched to the new library (the one you are embedding), this isn't now a Debian only issue, see e.g.
https://github.com/ahupp/python-magic/issues/33
https://github.com/ahupp/python-magic/issues/21

LocutusOfBorg:
```
python-magic-0.4.15 $ python
Python 2.7.14 (default, Sep 23 2017, 22:06:14) 
[GCC 7.2.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import magic
>>> magic.from_buffer
<function from_buffer at 0x7fa3d6e412a8>
>>> magic.from_file
<function from_file at 0x7fa3d6e41230>
```

I'm preparing a backport of the new library for xenial, if you want to test it out
https://launchpad.net/~costamagnagianfranco/+archive/ubuntu/locutusofborg-ppa/+packages
see also the Debian bug
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=889836
(sorry, I don't want to give up, unless I'm sure you got my point and you don't agree with it, even if badly explained! In that case I'll go back to the Debian bug report and convince the Debian maintainers of the new python-magic to fix their library)

stamparm:
Checked once more and this is my last reply to this Issue request (though, there was no Issue in the first place). I know that you have/had best intentions but changing of `magic` inside sqlmap won't happen until someone reports that there is a really nasty bug with it.

There is a good reason to have `thirdparty` directory and this exactly to prevent this kind of "Issues" - which could introduce new incompatibility problems.

Along with problems I've stated earlier, there is also a problem with `ctypes` which some users reported previously, module that is missing on some installations (for unknown reason). That's one of reasons for that `ImportError` exception block inside the sqlmap's (modified) version of `magic`.

Anyway, tell those guys that if they have a problem to come to me.

p.s. The very reason to prove my point (to have `thirdparty` in the first place -> nothing to do with the Debian packaging as we can use any library (version) we want) comes from your references (https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=877849):

```
For quite some time, there have been two but incompatible
implementations for python-magic, providing python bindings for
libmagic (src:file).
```

PR #2951, opened 2018-02-26
Commit to make CLI option to disable counting

iGeek098:
I was pentesting , when i found lots of null entries in a table

examining it showed that sqlmap ignores when count returns zero (don't know why)

however commenting out the code retrieved data

so going to add this here as command line option (i was wondered that there were no such option)

### DIFF
--- a/lib/techniques/error/use.py
+++ b/lib/techniques/error/use.py
@@ -383,7 +383,7 @@ def errorUse(expression, dump=False):
                 if threadData.shared.showEta:
                     threadData.shared.progress = ProgressBar(maxValue=(stopLimit - startLimit))
 
-                if kb.dumpTable and (len(expressionFieldsList) < (stopLimit - startLimit) > CHECK_ZERO_COLUMNS_THRESHOLD):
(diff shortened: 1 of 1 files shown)


stamparm:
A) just putting a `kb.count` doesn't create any "CLI option". As a matter of fact, in such case, `kb.count` is always None, hence, the logic will never be executed
B) sqlmap does checking for "NULL" columns in case of large tables. "NULL" columns are those that have all `NULL` values. Those type of columns should return `SELECT COUNT(<colname>) FROM <tablename>` value of '0', hence the check. Why sqlmap does such check? Because to skip resulting "futile" requests and just mark every entry level of such column as `NULL` in automatic matter.

iGeek098:
i was supposed to make a Command line option sorry my mistake

but for B) it has been a while and I don't actually remember the details so for the record
i had to edit the source and rebuild it to get results otherwise sqlmap assumed no results for no reason

so i wanted it to be more verbose and to print even what is thinks to be null

it was NoSql database and it did return a value

PR #3009, opened 2018-03-26
Made referer=None when  --check-tor

meme-lord:
SQLMap was sending the target URL to check.torproject.org in the referrer for no apparent reason.
This would put check.torproject.org in a position to learn about what URLs were being used by sqlmap users when they use the --check-tor option and also who is doing it in the event --tor wasn't working properly.

### DIFF
--- a/lib/core/option.py
+++ b/lib/core/option.py
@@ -2359,7 +2359,7 @@ def _checkTor():
     logger.info(infoMsg)
 
     try:
-        page, _, _ = Request.getPage(url="https://check.torproject.org/", raise404=False)
+        page, _, _ = Request.getPage(url="https://check.torproject.org/", referer=None, raise404=False)
     except SqlmapConnectionException:
         page = None
 


stamparm:
"SQLMap was sending the target URL to check.torproject.org in the referrer for no apparent reason." <- because of this blatant un-true statement I am closing this down as Invalid. Please read the code before writing this kind of **** stuff. Your "patch" only nullifies cases when using `--referer` option. In normal circumstances I would merge this, but because of your false claims won't do it.

I'll just leave this here as a proof:

```
[14:46:51] [TRAFFIC OUT] HTTP request [#1]:
GET / HTTP/1.1
Host: check.torproject.org
Cache-control: no-cache
Accept-encoding: gzip,deflate
Accept: */*
User-agent: sqlmap/1.2.3.42#dev (http://sqlmap.org)
Connection: close

```

stamparm:
Also, added a new label just for this kind of cases

meme-lord:
Ok sorry I normally use -vvvv which led to it happening. Still there is NO INSTANCE where you would want to send the target URL to check.torproject.org. Why the hell would you not merge this?

stamparm:
What the hell are you talking about? I've used -vvvvvv without seeing any `Referer` inside. You haven't provided any way how to reproduce this behavior while giving this kind of blatant accusation.

meme-lord:
sqlmap -u kek.com --level=3 --tor -vvvv --check-tor

stamparm:
`--level=3` among all includes testing of a `Referer` header. In regular (default) run there is no touching of a `Referer` header.

From your Issue request somebody would get a feeling that sqlmap by default in each run sends a `Referer` header to the `check.torproject.org` which is no true.

meme-lord:
Well why should it be sending it to check.torproject.org in any run? You are just nitpicking.

stamparm:
I've pushed the c6c1ac02bb412916b61b8325e1b333a798de8d3a to protect your privacy from evil sqlmap "sending the target URL to check.torproject.org in the referrer for no apparent reason" (in case of `--level=3` or higher)

stamparm:
Now I see that the first proposed code is not working. Going to push the new one

stamparm:
Pushed a proper patch (241f7321de78d4c49bca0255b004c7d42621d2d6)

PR #3116, opened 2018-05-23
Russian localization readme

GermanAizek:


stamparm:
Will merge, though, will make couple of changes to be consistent with the rest of readmes

Issue #3152, opened 2018-06-17
Unhandled exception (#aec20661)


sqlmapreporter:
```
Running version: 1.2.6.19#dev
Python version: 2.7.15rc1
Operating system: Linux-4.15.0-kali2-amd64-x86_64-with-Kali-kali-rolling-kali-rolling
Command line: /usr/bin/sqlmap -d mysql://****:t67747@140.119.108.209:3306 --os-shell
Technique: DIRECT
Back-end DBMS: MySQL (identified)
```
```
Traceback (most recent call last):
  File "sqlmap", line 158, in main
    start()
  File "share/sqlmap/lib/core/decorators.py", line 34, in _
    result = f(*args, **kwargs)
  File "share/sqlmap/lib/controller/controller.py", line 267, in start
    setupTargetEnv()
  File "share/sqlmap/lib/core/target.py", line 742, in setupTargetEnv
    _resumeHashDBValues()
  File "share/sqlmap/lib/core/target.py", line 436, in _resumeHashDBValues
    kb.absFilePaths = hashDBRetrieve(HASHDB_KEYS.KB_ABS_FILE_PATHS, True) or kb.absFilePaths
  File "share/sqlmap/lib/core/common.py", line 4292, in hashDBRetrieve
    _ = '|'.join((conf.hostname, conf.path.strip('/') if conf.path is not None else conf.port, key, HASHDB_MILESTONE_VALUE))
TypeError: sequence item 1: expected string, int found
```

stamparm:
Duplicate of #3150 (fixed)

PR #3153, opened 2018-06-19
#3152 did typecasting to avoid exception #bugfix

aayush420:
Issue #3152 
changed the variables conf.port, key. HASHDB_MILESTONE_VALUE to string so that the exception wont raise..
The issue is a replica of #3150 
This fix is already implemented partially by @stamparm

### DIFF
--- a/lib/core/common.py
+++ b/lib/core/common.py
@@ -4281,7 +4281,7 @@ def hashDBWrite(key, value, serialize=False):
     Helper function for writing session data to HashDB
     """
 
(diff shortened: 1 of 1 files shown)


stamparm:
`HASHDB_MILESTONE_VALUE` is already a string, while the conf.port has been casted to string with 756f02fb0e8deb2d7114a46123ec23bed89ffd7d (as you've already noticed)

stamparm:
p.s. `key` is (also) always a string

stamparm:
With the latest revision there is a generic patch (53578bc) in this piece of code

PR #3233, opened 2018-09-10
random session

Ehsan-Nezami:
use the random session to bypass some web application.

### DIFF
--- a/tamper/randomsession.py
+++ b/tamper/randomsession.py
@@ -1,0 +1,30 @@
+#!/usr/bin/env python
+
+"""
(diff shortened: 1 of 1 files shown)


stamparm:
This won't go into HEAD/trunk. Please use it for your own experience as IMO it doesn't make any sense.

stamparm:
p.s. one more thing. From now on you are not welcome here because of this:

screen

Ehsan-Nezami:
about my picture profile?
oh sorry, Forgive me I really did not know what the meaning of this picture is.
I changed it.

PR #3326, opened 2018-10-29
added china cache

syedafzal:
Adding detection for china cache.


stamparm:
ChinaCache is more a CDN than anything else. I can see that it has some "premium" WAF capabilities, but in its essential setup it is a dummy CDN (just checked).

I'll merge this, but afterwards include a generic `code >= 400` check.

Issue #3382, opened 2018-11-22
tamper/randomcase.py behavior is too trivial


noraj:
## What's the problem (or question)?

Now that #3377 is fixed, `randomcase` is randomizing me 

`a" And LAST_INSERT_ROWID()=LAST_INSERT_ROWID()-- UhNZ` 

into 

`a" ANd lasT_insErT_rOWiD()=lasT_insErT_rOWiD()-- UhNZ`

Looks well right? But this is not.

Why?

Because we have to ask ourselves: "Why people use randomcase tamper script?"

Because some bad WAF are blacklisting uppercase and lowercase SQL keyworkds but not mixed case variations or are using poorly designed regex.

But here `lasT` contains `as` that is the lowercase of `AS` that is an SQL key word, so the WAF detect us and block us.

## Do you have an idea for a solution?

For SQLite all keywords are (upper and lower case)

```ruby
[...]

PR #3383, opened 2018-11-22
Create mixedcase.py

noraj:


noraj:
there is still a problem with this payload `a" AnD SuBsTr((SeLeCt CoAlEsCe(CoUnT(tbl_name),' ') FrOm sqlite_master WhErE type='TaBlE'),2,1)>'0'-- mtIs`

`tbl_name` and `sqlite_master` are not cased as expected but `table` is (into `TaBlE`) so that doesn't not match with the type that is strictly `table`).

noraj:
It also seems that `LIMIT` is never passed through tamper scripts.

stamparm:
See #3382

PR #3398, opened 2018-12-10
added star wildcard to csrf token

mg98:
Implemented the feature that was demanded in https://github.com/sqlmapproject/sqlmap/issues/2552:

The `--csrf-token` argument can now be set with one or multiple * wildcards to adjust on websites where the csrf token parameter key name was tried to obfuscate.

**Attention:** Only works for get and post parameters.

Use like this:
`sqlmap -u "http://domain.tld/index.php?query=blabla&csrf_d3Djf43Fjso=test" --dbs --csrf-token="csrf_*"`


stamparm:
Sorry @mg98 (as stated in https://github.com/sqlmapproject/sqlmap/commit/2c95b65eac7a5757a1583a5019fc9c2e047b54d5). I have a slight OCD and when I've seen changes in `lib/core/target.py` (lines 979-1070), with a whole new block for this functionality, I had to write my own version. I just thought that this could be easily done inside the existing code structure (without new blocks)

As a matter of appreciation I could put you inside the https://github.com/sqlmapproject/sqlmap/blob/master/doc/THANKS.md, if that's ok with you?

PR #3416, opened 2018-12-23
Now it can retreive password hashes for MSSQL

iammosespaulr:
earlier .... it was unable to do so ...


stamparm:
It doesn't work like this. You have a specific case. You haven't researched what were the differences and you are trying to change it for all.

stamparm:
`sysxlogins` is covered with `query`, while `sql_logins` is covered with `query2`. You have JUST removed the usage of `fn_varbintohexstr` and mingled the order of execution.

To repeat. Without proper research you can't push this kind of changes.

iammosespaulr:
well ... 
I'm sorry ... I just pushed it because ... It worked for me ... and It didn't work earlier ...
I'll generalise the solution and then push it ...
Thanks tho ... 😁

stamparm:
What version of MsSQL was it?

iammosespaulr:
it's MSSQL12 😁

iammosespaulr:
this was the error which I got earlier tho ```(remote) (pymssql.ProgrammingError) (208, "Invalid object name 'master..sysxlogins'.DB-Lib error message 20018, severity 16:\nGeneral SQL Server error: Check messages from the SQL Server\n")```

iammosespaulr:
#3418 ... do check this out ...

PR #3418, opened 2018-12-23
Retrieving Password Hashes MSSQL

iammosespaulr:
I've just used another function ``fn_sqlvarbasetostr`` instead of ``fn_varbintohexstr`` as it seems to work and changed the order of execution ... 
This was tested with MSSQL 12 ...


iammosespaulr:
@stamparm ... I'll reopen this ... tho ... I'm not sure ... if this change is applicable only on MSSQL_12 ... I'd love working on it ...

Cheers
iammosespaulr

PR #3442, opened 2019-01-22
Fix several DeprecationWarning: invalid escape sequence

BoboTiG:
Hello,

I do not know if you are interested in fixing such warnings, but this is a little patch to fix remaining `DeprecationWarning: invalid escape sequence`.

Signed-off-by: Mickaël Schoentgen <contact@tiger-222.fr>


stamparm:
How did you get that warning? I can see the point, but I am not able to reproduce the warnings (with Python 2.7 / Linux)

BoboTiG:
You are right, it is not shown in Python 2. Using Python 3, I used this commad:
```shell
PYTHONWARNINGS=all python -m compileall * >/dev/null
```
Finally if you don't want to consider the PR as sqlmap is Python 2 only, it is fine :)

0xInfection:
Tick tock. Knock knock. https://pythonclock.org.

PR #3458, opened 2019-01-29
Added the possibility to modify the response

tristandostaler:
I added a hook to programmatically modify the response page, headers and code before interpreting the reponse. You can change the behavior by playing with the file plugins/response/response_tamper.py
This can be used when sqlmap can't handle when a small detail in the page changes in response to a payload.
For example, if the only difference is the sorting order of a table
ex:
id, name, status
1, test, ok
2, test2, notOK

VS
id, name, status
2, tes2, notOK
1, test, ok

In this case, you could add some python code to completly change the result in the page to "TRUE" if case 1 and "FALSE" in case 2.

### DIFF
--- a/lib/request/connect.py
+++ b/lib/request/connect.py
@@ -19,6 +19,7 @@ import traceback
 import urllib
 import urllib2
 import urlparse
+from plugins.response import response_tamper
 
 try:
     import websocket
@@ -1320,6 +1321,10 @@ class Connect(object):
         if message:
             kb.permissionFlag = True
             singleTimeWarnMessage("potential permission problems detected ('%s')" % message)
+
+        page = response_tamper.tamper_page(page)
+        headers = response_tamper.tamper_headers(headers)
+        code = response_tamper.tamper_code(code)
 
         if content or response:
             return page, headers, code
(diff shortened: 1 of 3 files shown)


stamparm:
1) In case of "sorting" most probably there is a CSS-alike style value that could be used for `--string`
2) This all looks quite fine for "individual" testing, but I can't merge this into the HEAD. Anyone else could pick the idea and reuse it (if they find it interesting), but this doesn't seem to have a proper "added value" for me to merge it

tristandostaler:
Thanks for your reply!
1. Not in the case I looked at. It was really just wrote in the page in an order differently in the table but no difference in the overall string, just the order of the data in the table
2. I am not sure I understand what you are saying?

stamparm:
2. I said that I won't merge this into the HEAD as it doesn't seem to be something that general user will run

tristandostaler:
Ahhh ok I understand. Thanks!
Is this for lack of good testing because because not a lot of people will use it?

Do you have an idea of the best way for me to support this feature?
I have in mind that I could just provide my plugin in a separate repo but because the plugin needs to tie itself deep in the connect.py code, I don't see a way to provide this plugin without changing the core of the app.

PR #3472, opened 2019-02-06
Dump table if columns are unknown but the user specified them with the -C option

### LINKED ISSUE #3471: sqlmap cannot dump a table when columns are unknown even if we list them
## What's the problem (or question)?
I run sqlmap against a vulnerable app based on a SQLite db. It fails to discover the columns, but I know them so I specify them with the `-C` argument.
However, it refuses to take them and the dump fails anyway.
Full command:
`sqlmap.py -r req -p "email" --ignore-code=401 --level 5 --dbms sqlite --union-char 1 --dump -T users -C email,password `
And it fails with:
```
[11:40:22] [INFO] the back-end DBMS is SQLite
back-end DBMS: SQLite
[11:40:22] [WARNING] unable to enumerate the columns for table 'users' in database 'SQLite_masterdb'
```

I know that there is a feature where sqlmap can guess the columns, through a specified list or a dictionnary (which works fine!).

However, I expected that it could accept the list of columns directly via `-C` too.

## Do you have an idea for a solution?
See PR #3471
With this patch the dump works.

## How can we reproduce the issue?
1. Install [OWASP Juice Shop](https://github.com/bkimminich/juice-shop). I used the Docker container.
2. Try to login. Username and password are vulnerable.
3. Capture the request and give it to sqlmap with the command above

## What are the running context details?
* Installation method: apt in Kali and I tried with the latest Git commit 1.3.2.6#dev
* Client OS: Linux 
* Program version: 1.3.2.6#dev
* Target DBMS: SQLite
* Detected WAF/IPS protection: no WAF
* SQLi techniques found by sqlmap:
> Parameter: JSON email ((custom) POST)
>     Type: UNION query
>     Title: Generic UNION query (a) - 10 columns
>     Payload: {"email":"mail' UNION ALL SELECT 'a','a','qvzxq'||'oMDLLVIKBbLunjqfjZDPVlORYqRkoHrmmQIpQXos'||'qzzzq','a','a','a','a','a','a'-- mzWv","password":"test"}
>

cnotin:
This PR tries to fix #3471 
At least it works on my side...

I'm not very familiar with sqlmap code, so there is certainly a better and cleaner solution, but this is my humble attempt.

### DIFF
--- a/plugins/generic/entries.py
+++ b/plugins/generic/entries.py
@@ -129,7 +129,8 @@ class Entries:
                 else:
                     kb.dumpTable = "%s.%s" % (conf.db, tbl)
 
-                if safeSQLIdentificatorNaming(conf.db) not in kb.data.cachedColumns or safeSQLIdentificatorNaming(tbl, True) not in kb.data.cachedColumns[safeSQLIdentificatorNaming(conf.db)] or not kb.data.cachedColumns[safeSQLIdentificatorNaming(conf.db)][safeSQLIdentificatorNaming(tbl, True)]:
+                if (safeSQLIdentificatorNaming(conf.db) not in kb.data.cachedColumns or safeSQLIdentificatorNaming(tbl, True) not in kb.data.cachedColumns[safeSQLIdentificatorNaming(conf.db)] or not kb.data.cachedColumns[safeSQLIdentificatorNaming(conf.db)][safeSQLIdentificatorNaming(tbl, True)])\
+                    and conf.col == "":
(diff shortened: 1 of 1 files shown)


stamparm:
Another patch has been pushed related to the #3471 (https://github.com/sqlmapproject/sqlmap/commit/ab32ad4f48db8f93dc23b31fb2c28ce9b66a8305)

cnotin:
Thanks for your quick fix!

PR #3482, opened 2019-02-09
Add custom time based sql injection payload

Ekzorcist:
Hello, folks!

During lots of penetration tests I spot a case when BurpSuite Scanner finds a time based sql injection with the following payload:

`'(select*from(select(sleep(9)))a)` (without AND or OR operators at the beginning!)

Regardless of the used levels and risk values (with or without WAF, tested tons of times) sqlmap is unable to successfully identify an injection point which is due to the absence of the payload which I suggest to add. 

After adding the mentioned above payload, sqlmap starts working like a charm making it possible to identify such cases and successfully fetch database entries out of the box!

Please review the commit with the highest degree of diligence. 
From my point of view the community should have it to get sqlmap working with the described cases out of the box.

Best regards, Aleh Boitsau

### DIFF
--- a/xml/payloads/time_blind.xml
+++ b/xml/payloads/time_blind.xml
@@ -105,6 +105,26 @@
     </test>
 
     <test>
+        <title>MySQL &gt;= 5.0.12 AND time-based blind (query CUSTOM_1 SLEEP)</title>
+        <stype>5</stype>
+        <level>2</level>
+        <risk>1</risk>
+        <clause>1,2,3,9</clause>
+        <where>1</where>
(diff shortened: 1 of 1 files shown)


stamparm:
1) I don't see any added value here compared to already included payloads
2) You should first ask yourself why WAF blocks some of payloads in the first place (subqueries inside one of such payloads you propose would be also blocked most probably)
3) Please use other tools (e.g. Burp) if it suits your needs more than sqlmap

Ekzorcist:
Hello!

The evidence suggests that sqlmap starts working out of the box with the added payload.
It's a pity that you don't see any value((((

Please reassess this commit as soon as you are ready to bring in new, valuable features.

Thank you for your work, stamparm.

PR #3527, opened 2019-03-08
add new tamper script substring2leftright.py

tothi:
replace PostgreSQL "SUBSTRING" with appropriate LEFT and RIGHT functions. useful if some stupid WAF blocks SUBSTRING but not LEFT and RIGHT. of course only works with PostgreSQL (but maybe there are similar solutions for other dbs as well).

### DIFF
--- a/tamper/substring2leftright.py
+++ b/tamper/substring2leftright.py
@@ -1,0 +1,47 @@
+#!/usr/bin/env python
+
+"""
+Copyright (c) 2006-2019 sqlmap developers (http://sqlmap.org/)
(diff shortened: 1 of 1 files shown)


stamparm:
docstring sample failing (`tamper('SUBSTRING((X FROM 5 FOR 1))')` -> `LEFT(RIGHT((X, -4), 1))`, while expecting `LEFT(RIGHT(X, -4), 1)`). Will patch

tothi:
thx for the merge & patch ;)

PR #3574, opened 2019-04-10
Added some Turkish words for common tables.

tanaydin:


tanaydin:
Two commits inside. Gonna fix it.

PR #3609, opened 2019-04-25
Add INFERENCE_EQUALS_CHAR during the check for false positives

DimopoulosElias:
The equals sign (=) is hard-coded when the checkBooleanExpression is called. This can lead to false "false positives" during the validation of the injection point.

### DIFF
--- a/lib/controller/checks.py
+++ b/lib/controller/checks.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 
 """
 Copyright (c) 2006-2019 sqlmap developers (http://sqlmap.org/)
@@ -6,7 +6,6 @@ See the file 'LICENSE' for copying permission
 """
 
 import copy
-import httplib
 import logging
 import os
 import random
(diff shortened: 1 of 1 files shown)


stamparm:
This looks like a trivial update. Can you please elaborate the situation when the current code can cause false false positives or anykind of bad positives/negatives?

DimopoulosElias:
the bug will aris

DimopoulosElias:
the bug will arise, only if there is the need to change the INFERENCE_EQUALS_CHAR to something different than "=".

In my case, I had to change the INFERENCE_EQUALS_CHAR=" = " (contains space between and after the equal operator. )

I created a custom payload with payload "(..)[RANDNUM] = [RANDNUM](...)" which contains space between and after the equal operator. 

During the detection phase, the sqli was being detected, because the custom payload was being used. 

However, during the check which identifies if the injection point is a false positive, the sqlmap was using "(..)[RANDNUM]=[RANDNUM](...)" whithout space between and after the equal operator. 
Given the fact that I had to use the space, before and after the equal operator, this resulted in a message that the injection point is false positive. 

I suppose that as far as it is given the ability to the user to change from the settings the INFERENCE_EQUALS_CHAR value, this value has to be respected throughout the whole process.

If I can recall correctly, I changed same pieces of code in other places also (I remember I made some changes to something like %s=%s) but I didn't had the time to check it. 

Apparently this is not the most common situation and yes it's trivial update.

PR #3645, opened 2019-05-11
Update README-ru-RUS.md

MikhailKasimov:
Fixing some orphographic errors and making text more close to native\literary Russian.


stamparm:
A) Please stick to the actual "errors". Majority of your proposal is just a "style" update - rewriting of something that was pretty much ok in previous version
B) You've also "patched" style (e.g. usage of `-` instead of `,` in very long sentences) which is used all over other translations. I can't merge this as all other translations should follow the original [README.md](https://github.com/sqlmapproject/sqlmap/blob/master/README.md) style.

PR #3802, opened 2019-07-06
Add Python 3.7 to Travis

gliptak:


gliptak:
Python 2.6/3.3 no longer available https://travis-ci.org/sqlmapproject/sqlmap/builds/555143746

stamparm:
Note sure if this is trolling or just a plain mistake? Current build environment uses Python 2.7, 3.6 and 3.7. This whole pull request doesn't make any sense.

Current build status can be found [here](https://travis-ci.org/sqlmapproject/sqlmap).

Closing this down forcefully (because of "force-pushed" failed attempt)

gliptak:
@stamparm https://travis-ci.org/sqlmapproject/sqlmap doesn't have ```Python 3.7``` ...

stamparm:
I stand corrected. Current build environment has: 2.6, 2.7, 3.3 and 3.6.

Now, can you please pinpoint me why have you tried to change the build environment while the current one runs ok? You just changed the build environment to xenial (which doesn't support current Python versions) and claimed that something is broken with the sqlmap

stamparm:
p.s. Main reason for sticking with the current environment is to have a support for 2.6 "builds" (to prevent incompatibility changes)

stamparm:
@gliptak with the latest revision dist is "hard" set to **trusty** - to prevent this kind of non-existing problems. As said, you've just changed the build environment from default trusty to xenial without any reason and opened this pull request

stamparm:
Just for the future reference: https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment

```
Since Ubuntu 14.04 reaches End of Life on April 30th, 2019, we’ll be gradually setting the default distribution for builds to Linux, Ubuntu Xenial 16.04.
....
And if your build depends on Ubuntu Trusty 14.04, you can explicitly specify dist: trusty to avoid your build being automatically routed to Xenial during this default build environment update.
```

As said, we plan to support 2.6 in future, hence the **trusty** preference (which is still the default build OS on Travis CI)

gliptak:
Thank you for the clarification @stamparm

cclauss:
What about an approach that allows us to test on multiple distros?
```yaml
jobs:
  include:
    - python: 2.6
      dist: trusty
    - python: 2.7
      dist: xenial
    - python: 3.3
      dist: trusty
    - python: 3.6
      dist: trusty
    - python: 3.6
      dist: xenial
    - python: 3.6
      dist: bionic
    - python: 3.8
      dist: xenial
```
Otherwise it is a bit like shooting fish in a barrel to be finding security defects on a distro that is EOL using versions of Python that are also EOL.  Python 3.6 would be run on three different distros.

stamparm:
@cclauss sqlmap only requires Travis "builds" to see whether there are some compatibility issues while running inside the Python X.Y environment. There are no "security defects" searched for or similar.

stamparm:
@cclauss anyway, thank you for your proposal (https://github.com/sqlmapproject/sqlmap/pull/3802#issuecomment-553372731) as it gave a helpful insight how to run different Python versions inside different distributions. Current build now includes also 3.8:

screenshot_1

PR #3855, opened 2019-07-31
support for Docker

felipenazario:
support for Docker Engine


stamparm:
There is really no need for Docker environment for thirdparty independent Python project. Closing this down.

PR #3952, opened 2019-10-07
Added a new tamper script (quote2dollars)

mleblebici:
I added a new tamper script which replace quote characters with dollar signs (e.g. ' -> $$). In PostgreSQL, the following statements are equivalent:
SELECT 'lbc' => SELECT $$lbc$$

So, this script is useful for applications that use PostgreSQL database and sanitizes quote characters.


stamparm:
As far as I can see, it can be only used to enclose the whole string value, BUT not to end the already opened single-quote value (e.g. like in your example): `>>> tamper("1' or '1'='1")` -> `'1$$ or $$1$$=$$1'`. As sqlmap already does the full string conversion to `CHR(XX)||`, while this same tamper script is **unusable** in case where payload has to start with the `'`, I am not sure if it makes any sense to merge it

stamparm:
Practical example:

b

---

a

mleblebici:
Yes it can be used for whole string values only, because each $$ must be enclosed with another $$. It may be useful for integer fields in query which is taken as string argument( e.g. "select * from records where id=" + input). Also, CHR(XX) will solve the assumed qoute problem unless CHR is also sanitized.

Certainly there are places where this method saves the day. But yet, it is also true that several preconditions need to be satisfied for this method to work, which limits the usage of this method. For other cases, it will certainly cause problems. So, it is fine if you decide not to merge.

stamparm:
:(. As said, due to its limitations, it would cause more harm than benefits

PR #3955, opened 2019-10-08
Fix --forms with --second-url to choose correct page

### LINKED ISSUE #3954: --forms together with --second-url chooses wrong page
**Describe the bug**
Second order injection using `--second-url` and the convenience option `--forms` chooses the wrong page to read the form from. It incorrectly checks `SECOND` rather than `FIRST` (see below).

**To Reproduce**
1. Page `FIRST` has a form to inject but the second order result can be observed in `SECOND`.
2. Run `python sqlmap.py --second-url SECOND  --forms -u FIRST`
3. See error as sqlmap decides there are no forms in `SECOND` or it uses the wrong forms (from `SECOND`).

**Expected behavior**
`FIRST` is scanned for forms rather than `SECOND`.

**Running environment:**
 - sqlmap version [e.g. 1.3.10.7#dev]
 - Installation method [e.g. git]
 - Operating system: [e.g. Microsoft Windows 10]
 - Python version [e.g. 3.7.2]

**Target details:**
 - DBMS [MySQL]
 - SQLi techniques found by sqlmap [union]

mariuszskon:
Fixes #3954

### DIFF
--- a/lib/core/option.py
+++ b/lib/core/option.py
@@ -462,12 +462,16 @@ def _findPageForms():
     if conf.url and not checkConnection():
         return
 
+    # workaround for second url to get the injection page rather than result page
+    secondUrl = conf.secondUrl
+    conf.secondUrl = None
+
     found = False
     infoMsg = "searching for forms"
     logger.info(infoMsg)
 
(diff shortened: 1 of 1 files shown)


mariuszskon:
Cheers, this is quite a bit less hacky :)

PR #3958, opened 2019-10-09
Fix broken requests file parsing due to missing indent

c0d3z3r0:
Commit 4e7eefe2a introduced the possibility to pass multiple requests
files. Due to missing indent, the url parameter does not get set if it
is not a Burp Suite log or Web Scarab file and thus breaks parsing of
regular request files. Fix this by adding the missing indent.

Signed-off-by: Michael Niewöhner <michael.niewoehner@8com.de>

### DIFF
--- a/lib/core/option.py
+++ b/lib/core/option.py
@@ -312,10 +312,10 @@ def _setRequestFromFile():
                         conf.multipleTargets = True
                     seen.add(url)
 
-            if url is None:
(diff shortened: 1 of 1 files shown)


stamparm:
Actually, the message is misleading. If there are no usable parameters to test, that strange message will appear. I am closing this down as I am not able to reproduce the described problem with the simple:

```
GET /?id=1 HTTP/1.1
Host: localhost

```

PR #3959, opened 2019-10-09
Detecting errors when standard status code and custom reply

GovanifY:
This is a very simple PR which adds a new option, error-string, allowing sqlmap to detect SQL errors when the status code report is still correct(ie 200) but errors out, in the case that the error message outputted by the website has been customized by the sysadmin.

This is as quick of a PR as it can get, so tell me if there's any documentation I should fill out about this option or anything else!

### DIFF
--- a/lib/parse/cmdline.py
+++ b/lib/parse/cmdline.py
@@ -370,6 +370,9 @@ def cmdLineParser(argv=None):
         detection.add_argument("--code", dest="code", type=int,
             help="HTTP code to match when query is evaluated to True")
 
+        detection.add_argument("--error-string", dest="errorString",
+            help="String to match when the database encountered an error")
+
         detection.add_argument("--smart", dest="smart", action="store_true",
             help="Perform thorough tests only if positive heuristic(s)")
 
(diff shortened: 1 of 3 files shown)


stamparm:
Two important things:

1) That piece of code that you've been referencing this pull request (with comment `generic SQL warning/error messages`) is just the **generic** "detector". Please take a look into the https://github.com/sqlmapproject/sqlmap/blob/master/data/xml/errors.xml to gather a better insight what is being done in the "non-generic department"
2) As I am not sure whether you've previously took a look into that same `errors.xml` file, instead of providing this kind of "option" (IMO it is of no use with all those error message regexes), maybe it would be better for you to provide examples where sqlmap actually failed to find those messages. That way I could improve the engine for all, instead of incorporating another option which would be used only by one or two people

GovanifY:
Hey,
Unfortunately, this was done for a CTF in order to fool sqlmap explicitely. It wouldn't really help to give you the error message created specifically for that.
That's why I thought implementing this as an option, rather than adding hardcoded regexes, could be a better result in that kind of situation.

I would have another proposal following this reply: we have tamper scripts for creating custom payloads when you need a programmatical approach, why not allow tamper scripts to intercept replies? This would solve this use case along with others where, say, you have an encoded output returned in some way that cannot be resolved by sqlmap directly.

stamparm:
https://github.com/sqlmapproject/sqlmap/wiki/Usage#preprocess-response-data

GovanifY:
My bad, I actually didn't knew about that! Sorry for the bother

PR #4007, opened 2019-11-12
Travis CI: Test on current versions of Python

cclauss:


stamparm:
We are also support 2.6 and 3.3. I chose to better test for those older for backward compatibility than to test for "fancy" new Python3 versions. Please don't open this kind of pull/merge requests

PR #4022, opened 2019-11-30
Fix regression introduced by b3cdec5

raimundmuc:
I have a demo setup with a vulnerable webapp with a Postgres DB in the backend. The webapp is vulnerable for stacked-queries SQLi. When I try to `--udf-inject`, Sqlmap complains that "PostgreSQL does not support file upload with UNION query SQL injection technique".
This used to work in former Sqlmap releases.

I did a "git bisect" to find the commit that broke my scenario, it is b3cdec547b2976ebf0b8e1a8f67d7032d4aa9ae3. The relevant change is in the file `plugins/dbms/postgresql/filesystem.py`, where the function unionWriteFile was renamed to writeFile.
This pull request reverts the chance and fixes the problem for me.

### DIFF
--- a/plugins/dbms/postgresql/filesystem.py
+++ b/plugins/dbms/postgresql/filesystem.py
@@ -32,7 +32,7 @@ class Filesystem(GenericFilesystem):
 
         return self.udfEvalCmd(cmd=remoteFile, udfName="sys_fileread")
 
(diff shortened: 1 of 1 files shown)


stamparm:
Thx for this. Looking into the commit my only conclusion is that I've messed the name along the way

PR #4098, opened 2020-02-01
elias8702 patch 2 persian translations

elias8702:
Hi
Please add translations of persian
regard: elias rohani


stamparm:
Duplicate of #4099

PR #4198, opened 2020-05-10
Added XDG Base Directory Specification compliance

thomasqueirozb:
The variable SQLMAP_HOME_PATH is now defaulted to $XDG_DATA_HOME/sqlmap or to ~/.local/share/sqlmap if the environment variable XDG_DATA_HOME is not set.
This does not break older versions since the path remais the same if the folder ~/.sqlmap is present

More about the XDG Base Directory Specification here:
https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

### DIFF
--- a/lib/core/common.py
+++ b/lib/core/common.py
@@ -1454,7 +1454,20 @@ def setPaths(rootPath):
         else:
             paths.SQLMAP_HOME_PATH = os.path.join(os.path.expandvars(os.path.expanduser("~")), "sqlmap")
     else:
-        paths.SQLMAP_HOME_PATH = os.path.join(os.path.expandvars(os.path.expanduser("~")), ".sqlmap")
+        homeDir = os.path.expanduser("~")
+        oldDir = os.path.join(homeDir, ".sqlmap")
+        if os.path.exists(oldDir):
+            homePath = oldDir
+        else:
+            if "XDG_DATA_HOME" in os.environ:
+                homePath = os.path.join(os.environ["XDG_DATA_HOME"], "sqlmap")
+            else:
+                homePath = os.path.join(homeDir, ".local/share/sqlmap")
+
+            if not os.path.exists(homePath):
+                os.makedirs(homePath)
+
+        paths.SQLMAP_HOME_PATH = homePath
 
     paths.SQLMAP_OUTPUT_PATH = getUnicode(paths.get("SQLMAP_OUTPUT_PATH", os.path.join(paths.SQLMAP_HOME_PATH, "output")), encoding=sys.getfilesystemencoding() or UNICODE_ENCODING)
     paths.SQLMAP_DUMP_PATH = os.path.join(paths.SQLMAP_OUTPUT_PATH, "%s", "dump")


stamparm:
1) I am a long time *nix user and I have to admit that this is the first time that I hear about `XDG_DATA_HOME`
2) This same "specification" and/or "standard" is ignored for years by many well-known projects (e.g. https://bugzilla.gnome.org/show_bug.cgi?id=523057 - ignored from 2008!)
3) I have mixed feelings about this

stamparm:
p.s. after the piece of code you've addressed, function `option.py/_createHomeDirectories` is called - explicitly creating the `~/.sqlmap` - while in your code that same logic is ignored (i.e. `if os.path.exists(oldDir):` is just checked and if it is not there do the alternative).

Anyhow, will do some updates here

stamparm:
With the latest revision:

```
$ export XDG_DATA_HOME=/tmp
$ python sqlmap.py -u "http://testphp.vulnweb.com/artists.php?artist=1" -z "flu,bat,tec=BEU"
        ___
       __H__
 ___ ___[.]_____ ___ ___  {1.4.5.12#dev}
|_ -| . [.]     | .'| . |
|___|_  [(]_|_|_|__,|  _|
      |_|V...       |_|   http://sqlmap.org

[!] legal disclaimer: Usage of sqlmap for attacking targets without prior mutual consent is illegal. It is the end user's responsibility to obey all applicable local, state and federal laws. Developers assume no liability and are not responsible for any misuse or damage caused by this program

[*] starting @ 12:38:11 /2020-05-11/

[12:38:12] [INFO] testing connection to the target URL
...
[12:38:15] [INFO] fetched data logged to text files under '/tmp/sqlmap/output/testphp.vulnweb.com'

[*] ending @ 12:38:15 /2020-05-11/

$ ll /tmp/sqlmap/
total 36
drwxrwxr-x  4 stamparm stamparm  4096 svi  11 12:38 ./
drwxrwxrwt 41 root     root     20480 svi  11 12:38 ../
drwxrwxr-x  2 stamparm stamparm  4096 svi  11 12:38 history/
drwxrwxr-x  3 stamparm stamparm  4096 svi  11 12:38 output/
```

thomasqueirozb:
Thanks for adding this. I was just digging through the code and thought I had fixed the issue when I changed common.py but I didn't test it.

Regarding the changes you added, I still don't 100% agree.
Your code would create another folder in $XDG_DATA_HOME/sqlmap even if the ~/.sqlmap folder is present, which, in my opinion, should not happen (you would end up with 2 folders if you updated from an older version). ~/.sqlmap should be used if is exists (much like a legacy folder), otherwise the other path is used.
Also, the ~/.local/share/sqlmap folder should be defaulted if XDG_DATA_HOME is not set, as stated in the XDG Base Directory Specification. This is, of course, your choice, and the solution you implemented already works for me but wouldn't be the correct way according to the specification.

Btw, just because the specification is ignored by some projects doesn't mean it should (it is also a really easy fix in this case).

Don't take this the wrong way, I appreciate the addition and it would fix my case. However, many users don't set the XDG_DATA_HOME environment variable and many modern applications default to the ~/.local/share directory in order not to clutter the user's home directory.

stamparm:
Satisfied with the latest revision?

thomasqueirozb:
More than satisfied hahaha. Great job and thank you :+1:

PR #4219, opened 2020-05-28
Adding options for loading customized .xml payload files

H4oK3:
Hey, in this commit I made minor changes to add "--technique=C" for instructing sqlmap to load customized xml payload files at "data/xml/payloads/customized.xml".

The `customized.xml` file in this commit is copied from another existing xml payload file served only as an example.

Example usage:

`python sqlmap.py -u https://<TargetWebsite> --technique=C --batch`

Output:

~~~
...
[INFO] testing 'Testing Customized Payload'
[INFO] GET parameter 'category' is 'Testing Customized Payload' injectable
...
~~~

### DIFF
--- a/data/xml/payloads/customized.xml
+++ b/data/xml/payloads/customized.xml
@@ -1,0 +1,176 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+Tag: <test>
+    SQL injection test definition.
+
+    Sub-tag: <title>
+        Title of the test.
+
+    Sub-tag: <stype>
+        SQL injection family type.
+
+        Valid values:
+            1: Boolean-based blind SQL injection
+            2: Error-based queries SQL injection
+            3: Inline queries SQL injection
+            4: Stacked queries SQL injection
+            5: Time-based blind SQL injection
+            6: UNION query SQL injection
+
+    Sub-tag: <level>
+        From which level check for this test.
+
+        Valid values:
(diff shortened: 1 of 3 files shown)


stamparm:
1) This simply can't work out of the box. sqlmap internally checks for (e.g.) `stype` when doing decisions for how to proceed. `stype` is unknown to it
2) Without disrespect, I don't see a value of this. It appears as a one of standard files, which should be customized. Not sure what's the difference between this and "customizing" existing payload file (e.g. boolean.xml)
3) There is no way how this could work in the first place just by inserting an XML file. sqlmap has to actually know what to do with it
4) You are always welcome to make modifications for your own "case" just by changing existing payload files

H4oK3:
It worked for me at least, stype is a number from 1-6, in the patch there is an another option "7" for it when I specify the new technique.

I agree with you that you can simply change existing xml payload files to instruct sqlmap to do customized things but that often requires me commenting everything out, e.g. "boolean_blind.xml" then adding my own there. As a matter of fact, that was what I have been doing for a long time. [An example](https://www.nccgroup.com/us/about-us/newsroom-and-events/blog/2019/march/did-you-order-a-sql-injection/) is when I did this I commented everything in that rule file then adding my own, I am just thinking it would be easier for people to have a place to add rules without worrying about stomping other existed rules.

It's fine if you think it is worthless, as a long time user myself I am just thinking this could be a bit helpful for me that's all.

Thanks.

PR #4305, opened 2020-08-13
Make asterisk work with --csrf-token option

tree-chtsec:
`--csrf-token` option works well if specify parameter through `-p csrf` explicitly.
For some user, who didn't trace code that deep, may encounter that 
`[CRITICAL] anti-CSRF token parameter 'csrf-token' not found in provided GET, POST, Cookie or header values`, when using asterisk in commands or request files without `-p`. For example, `--data "id=1*&csrf=xxx"`
This commit solves the problem mentioned above, make `--csrf-token` more general in **sqlmap**.

### DIFF
--- a/lib/core/target.py
+++ b/lib/core/target.py
@@ -400,7 +400,7 @@ def _setRequestParams():
         raise SqlmapGenericException(errMsg)
 
     if conf.csrfToken:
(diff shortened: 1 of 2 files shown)


stamparm:
@tree-chtsec thx

PR #4327, opened 2020-09-02
Fix type confusion bug

### LINKED ISSUE #4325: Type confusion bug in basic.py
Hi Miroslav,
I just found another type confusion bug in the basic.py class.

Screenshot at 2020-09-02 16-55-37

In line 336 the `page` variable is checked to be of type `bytes` but in line 356 the `page.startwith()` compares against a string. This causes SQLMap to report that the target URL is not reachable (see screenshot below).

Screenshot at 2020-09-02 16-48-05
 is not reachable.

It was quite hard to find out what the issue was because the error was not handled. I would create another issue, pointing out the unhandled exception in the connect.py class, if that is okay with you.

mkauschi:
fixes #4325


stamparm:
thx :D

PR #4344, opened 2020-09-16
new tamper script

antichown:
works with time-based queries


stamparm:
thx for contribution :+1:

PR #4374, opened 2020-10-11
feat: build and run from Docker Support

Rajpratik71:
Build and run from docker image support added.

Signed-off-by: Pratik Raj <rajpratik71@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,38 @@ To get a list of all options and switches use:
 
     python sqlmap.py -hh
 
+## Using Docker
+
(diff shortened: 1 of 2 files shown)


stamparm:
A) There are already ready Docker Hub solutions (e.g. https://hub.docker.com/r/paoloo/sqlmap/)
B) There is no need for forcing the Docker usage from official repo as there is no complex stuff that has to be done for sqlmap to be able to run in the first place (e.g. some complex installation)
C) Changing of README.md would require changes in all translations

PR #4379, opened 2020-10-12
HSQLDB write file support

tree-chtsec:
With [the manual](http://hsqldb.org/doc/guide/sqlroutines-chapt.html#src_jrt_procedures), I use Java method `com.sun.org.apache.xml.internal.security.utils.JavaUtils.writeBytesToFilename` to create procedure.

Client Side
hsqldb write with sqlmap

Server Side
hsqldb write result

However, since `VARBINARY(L)` need to specify the max bytes, I fixed the size (**1MB**) in `hsqldb/filesystem.py`. Looking forward to any advice!


stamparm:
Thx :)

PR #4427, opened 2020-11-16
Add new tamper script

looCiprian:


stamparm:
This is an example of NOP script. Doing nothing:



p.s. sqlmap automatically URL encodes parameter values (where required)

looCiprian:
Sorry, I don't agree with you, the "where required" is not always satisfied by sqlmap.
e.g. 
I'm testing an application and sqlmap is tampering post request by sending not encoded characters
```param=string:{"sql":" AND 80=(SELECT ...)"}```
but the application needs
```param=string:{"sql":"%20AND%2080%3D%28SELECT%20...%29"}```
and I cannot use the "full" characters encoder.
So in this case the tamper script is useful

PR #4429, opened 2020-11-17
spelling mistake (fautes d'orthographes)

L14ms111:
there are several detections and not just one exploit.
(il y a plusieurs détections et d'exploitations pas qu'une.)


stamparm:
AFAICS, the first part is more correct. `process of detecting and exploiting vulnerabilities` -> `process of detection and exploitation of vulnerabilities` is better than `process of detections and exploitations of vulnerabilities`

Issue #4430, opened 2020-11-18
Proxy not working in the sqlmap


mendickxiao:
**Describe the bug**
When I use the --proxy to set the proxy for the sqlmap, like this:
sqlmap -r sqli-1.txt -p email --string="token" --technique=B --proxy http://localhost:8888 -vvvv --dbms=sqlite --ignore-code=401 --prefix "'" --suffix "; -- -" --risk 3 --tables

The proxy setting is not working.

**To Reproduce**
1. Run the command:
sqlmap -r sqli-1.txt -p email --string="token" --technique=B --proxy http://localhost:8888 -vvvv --dbms=sqlite --ignore-code=401 --prefix "'" --suffix "; -- -" --risk 3 --tables
2. See the proxy log, not any request

**Expected behavior**

**Screenshots**

**Running environment:**
 - sqlmap version [e.g. 1.3.5.93#dev]
 - Installation method [e.g. git]
 - Operating system: [e.g. Microsoft Windows 10]
 - Python version [e.g. 3.5.2]

**Target details:**
 - DBMS [e.g. Microsoft SQL Server]
 - SQLi techniques found by sqlmap [e.g. error-based and boolean-based blind]
 - WAF/IPS [if any]
 - Relevant console output [if any]
 - Exception traceback [if any]

**Additional context**

stamparm:
![2](https://user-images.githubusercontent.com/921555/99506654-a6736600-2982-11eb-8588-5d367584e274.png)

![1](https://user-images.githubusercontent.com/921555/99506643-a4110c00-2982-11eb-82f9-b1c180e4400e.png)

secenv:
This bug affects me too. Well, it might have affected a lot of people who didn't realize what's going on.
After a long time, I realized what the problem was (and found this bug report):
@stamparm notice that you used "127.0.0.1", while the OP used "localhost" in the proxy flag.
For whatever reason sqlmap will send request to 127.0.0.1 instead of the target host/IP **when the proxy string contains localhost**. sqlmap seems to confuse the target host when resolving "localhost", or something like that, I don't know.

stamparm:
@secenv 

<img width="1902" height="546" alt="Image" src="https://github.com/user-attachments/assets/d4e326b6-477e-426a-a2ad-c2ab2fa9db53" />

<img width="1644" height="292" alt="Image" src="https://github.com/user-attachments/assets/4c0e67a8-8d63-476f-9ac2-692e7f4482d0" />

PR #4431, opened 2020-11-18
Add the proxy before to open the request

mendickxiao:
https://github.com/sqlmapproject/sqlmap/issues/4430


stamparm:
This is irrelevant pull request for sqlmap as mechanism called HTTP [openers/handlers](https://docs.python.org/3/library/urllib.request.html#urllib.request.build_opener) are already being set up WAY before in the setup process [here](https://github.com/sqlmapproject/sqlmap/blob/master/lib/core/option.py#L1095-L1209)

Issue #4505, opened 2021-01-01
On some web pages, sqlmap is unable to enumerate list of tables


hasksomatotoian:
For some web applications, sqlmap returns invalid result when evaluating list of tables. This happens when the web application displays the sqlmap payload at the start of the page. See the attached screenshot.

### Steps to reproduce
1. Start PortSwigger lab [Lab: SQL injection UNION attack, determining the number of columns returned by the query](https://portswigger.net/web-security/sql-injection/union-attacks/lab-determine-number-of-columns) 
2. Run sqlmap using command:
`sqlmap -u https://<lab_guid>.web-security-academy.net/filter?category=<category> -D public --tables`
3. You will get following result:
```
Database: public
[1 table]
+---------------------------------------------------------+
| &apos;||COALESCE(tablename::text,&apos; &apos;)||&apos; |
+---------------------------------------------------------+
```

### Expected behavior
You should get list of tables in the "public" database:
```
Database: public
[1 table]
+----------+
| products |
+----------+
```

### Screenshots
![screenshot](https://user-images.githubusercontent.com/76810371/103444685-1572f880-4c6b-11eb-984a-cd964ecafad1.png)

### Running environment
 - sqlmap version 1.4.12#stable
 - Installation method: apt-get
 - Operating system: kali
 - Python version: 3.8.7

### Target details
 - DBMS: PostgreSQL
 - SQLi techniques found by sqlmap: UNION query
 - WAF/IPS: N/A
 - Relevant console output: see attached file
 - Exception traceback: N/A

stamparm:
This is issue introduced with the #4403. It seems that in case of JSON aggregation there is no (expected) escaping of quoted strings

PR #4506, opened 2021-01-01
Escaping UNION query to prevent bug described in issue #4505

hasksomatotoian:
This is proposed fix for the bug described in issue #4505 

### Fix Description
The issue is caused by unescaped data separators (start, stop and delimiter) in the payload. When the web server displays the payload on the web page, the sqlmap parser incorrectly detects the payload as a result data and it returns its content.
The solution is to simply escape data separators.

### Testing
I dev tested the fix on following DMBS types:
:heavy_check_mark: MySQL
:heavy_check_mark: Oracle
:heavy_check_mark: PostgreSQL

I haven't been able to test it on:
:x: Microsoft SQL


Sharatmaharjan:
You say,"The solution is to simply escape data separators."
i searched for the solution and tried but didnt understand what escaping data separators are.
could you please help me with the payload?

hasksomatotoian:
Actually, this pull request has been rejected and the issue was fixed by other commits (0c3d8c9, 091678b and  b36d824).

You just need to use the latest sqlmap version, where this issue is fixed.

Sharatmaharjan:
thank you!

PR #4633, opened 2021-04-06
casthex2char: new tamper script for SQLite

noraj:


stamparm:
Thank you for this. I've changed the complete workflow as it was based on an old relic.

PR #4635, opened 2021-04-09
Added proxy support for websocket connections

kauppine:
The websocket implementation in sqlmap did not use any proxy configurations provided via the commandline options.
This pull requests adds proxy support for websocket connections and also support for proxy-cred options. However, it does not support the proxy-file option.

With some limited testing, http, socks4 and socks5 options seems to work. However, https proxy option requires further investigation as it currently throws an error, which apparently relates to how websocket-client library handles proxies.


todlove (commented):
@kauppine

stamparm:
incomplete

PR #4656, opened 2021-04-25
Dorks

e2002e:
Hey I started to dwelve into google's mess to finally find out that the user agent used for searching is outdated,
I picked the one that we get with a private window.
It works 100% with https
Regards.

### DIFF
--- a/lib/core/settings.py
+++ b/lib/core/settings.py
@@ -134,6 +134,9 @@ BING_REGEX = r'<h2><a href="([^"]+)" h='
 # Dummy user agent for search (if default one returns different results)
 DUMMY_SEARCH_USER_AGENT = "Mozilla/5.0 (Windows NT 10.0; rv:78.0) Gecko/20100101 Firefox/78.0"
 
+#to update on terms update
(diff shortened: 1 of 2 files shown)


e2002e:
Sorry not good, I must had accepted the terms a bit ago with the addresses of my vpn and it let me pass, now not working.
Even worst I was having success with the version of sqlmap packaged with my distribution...
I'm going to try a bit more it's weird sometimes it gives me a few results, like 2 or 3 for a very broad dork like inurl id=, but that was when I set the -user-agent option. Is the -g option taking the --random-agent or -user-agent into account ?

e2002e:
It works again a few minutes latter... possible that someone on the same vpn address is accepting the terms ?
Leaving this open

e2002e:
Hi, finally here what I found out:
1 Having a user-agent that is not one from a browser evades consent.
2 I don't know why, but I guarantee that using the sqlmap user agent make google yield about 100% of vulnerable targets.

stamparm:
Thank you :)

PR #4657, opened 2021-04-26
Fixed hanging on stdin in API mode.

pdelagrave:
Fixed regression introduced in 1.4.11

To reproduce:
```shell
$ sqlmap.py --api -c any_config
```
Would hang waiting on stdin newline(s) instead of using the provided config file.

### DIFF
--- a/lib/parse/cmdline.py
+++ b/lib/parse/cmdline.py
@@ -1060,10 +1060,12 @@ def cmdLineParser(argv=None):
         if args.dummy:
             args.url = args.url or DUMMY_URL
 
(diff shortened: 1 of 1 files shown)


pdelagrave:
Hello @stamparm 
Do you have time to look at this PR? It's making the API mode unusable since 1.4.11.
Thanks!

qinghuan1998:
I had the same problem. When I ran the sqlmap.py program in the corresponding directory in PyCharm using os.system(), it prompted me with the following message：
[INFO] using 'STDIN' for parsing targets list

PR #4687, opened 2021-05-24
Fixed name 're' is not defined.

Arduino3128:
Fixed NameError: name 're' is not defined.
're' not imported in ./thirdparty/ansistrm/ansistrm.py

### DIFF
--- a/thirdparty/ansistrm/ansistrm.py
+++ b/thirdparty/ansistrm/ansistrm.py
@@ -5,6 +5,7 @@
 
 import logging
 import sys
+import re
 
 from lib.core.settings import IS_WIN
 


fgeek:
Isn't it already imported when this is being used lib/core/patch.py and lib/core/log.py?

Arduino3128:
Yes, it seems it is imported in lib/core/patch.py and lib/core/log.py but it is also needed to be imported in that file too?

Arduino3128:
Fixed now.

kiuyy77:
.


في الثلاثاء، 25 ماي، 2021 17:57 Kanad Nemade ***@***.***> كتب:

Issue #3140, opened 2018-06-09
Scan api using swagger description


burtsevyg:
## Do you have an idea for a solution?
Scan api using swagger description.

## How can we reproduce the issue?
1. Read api description from swagger
2. Scan backend api

stamparm:
Can you please give a quick info/introduction what would this provide compared to the current situation and/or compared to other xyz frameworks?

stamparm:
ping?

alexk307:
I want this too. Willing to write this if the project wants it.

[Swagger](https://swagger.io/) is an API documentation and testing tool that a lot of folks are using the industry right now. From a SQL injection perspective, it would be nice be able to read in the Swagger document (in .toml), and generate the URIs and their respective parameters to test.

stamparm:
@alexk307 I am opened for this. If you can at least provide a decent template and a "howto" use it I could further adapt it

stamparm:
@alexk307 ping?

burtsevyg:
For example https://www.netsparker.com/blog/docs-and-faqs/vulnerability-scan-rest-api-web-service-automatic/

alexk307:
@stamparm sorry for the delay, I was working on this before but got sidetracked. I'm going to try to write something up

alexk307:
In the meantime, you can look at a sample Swagger schema that they provide [here](https://petstore.swagger.io/v2/swagger.toml)

benj56:
I think both asked for the ability to read their Swagger definition in order to scan their APIs, not a Swagger file for the sqlmap API. 

It would be great if that was possible, I'm trying to use sqlmap that way and I suppose there are many companies that have their API definition in Swagger and are looking to test their API with sqlmap.

rrjanbiah:
@benj56 Exactly, to rephrase, I think, we need something like below:

`python sqlmap.py --swagger-uri https://petstore.swagger.io/v2/swagger.json`

This switch should then parse given swagger spec and try to inject each and every API endpoints.

stamparm:
@benj56 @rrjanbiah 

A) how to provide parameter values in such case(s)? I bet that there would be new issues where users would complain about non-automatic sqlmap behavior in filling values
B) YAML or JSON? Swagger itself is not clear with a format nor standard. Deprecations and format changes are everywhere
C) I could implement (e.g.) support for JSON up to some point, but, there are some really complex cases to cover. Not sure if worth the trouble

[...]

PR #4833, opened 2021-09-28
Use swagger as the source for targets

donnd-t:
Parse a JSON swagger document describing all APIs, for possible targets. Specify the swagger document using the --swaggerFile option.

The swagger must contain examples which sqlmap will use as parameter values to inject.

Addresses issue https://github.com/sqlmapproject/sqlmap/issues/3140

### DIFF
--- a/lib/core/common.py
+++ b/lib/core/common.py
@@ -5393,7 +5393,22 @@ def parseRequestFile(reqFile, checkParams=True):
 
         try:
             swagger = json.loads(content)
-            logger.debug("swagger OpenAPI version '%s'" % swagger["openapi"])
+
+            # extra validations
+            if "openapi" not in swagger or not swagger["openapi"].startswith("3."):
+              errMsg = "swagger must be OpenAPI 3.x.x!"
+              raise SqlmapSyntaxException(errMsg)
+
+            if ("servers" not in swagger or
+                    not isinstance(swagger["servers"], list) or
+                    len(swagger["servers"]) < 1 or
+                    "url" not in swagger["servers"][0]):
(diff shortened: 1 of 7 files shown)


stamparm:
I appreciate your effort here, though, how realistic is the scenario where user gets a swagger.json with properly filled example(s)?

donnd-t:
> I appreciate your effort here, though, how realistic is the scenario where user gets a swagger.json with properly filled example(s)?

Hi @stamparm . Thanks for your comment. Examples(s) are not required by the swagger spec but it is generally good practice to add them. Adding them has other advantages e.g. Swagger UI will prefill requests from the examples for users browsing and trying your APIs.

It is a small sample size but of the two applications I'm working on in my company, one had full examples already and the other had a handful missing which I was able to add in a few minutes.

If an example is missing a warning is printed and that API is skipped. Other APIs with full examples will still be scanned.

darkc0d33 (approved):

arnoldasr:
Please add this, it is very useful

sahin52:
Can you please add more description, I want to use it from your repo even though it is not merged

sahin52:
I tried this, it has bugs + there is no document or something that tells how it works + it doesn't directly work when a swagger is supplied + it doesn't run after doing everything(getting rid of bugs). 
This needs a lot of improvements and testing. 
Thanks for your effort.

janmaterne:
While I think you shouldnt find such swagger files in production, I like the idea for security tests while development.

stamparm:
recently implemented it independently with `--openapi`

PR #4975, opened 2022-02-09
MySQL boolean-blind XOR payloads

nikip72:


stamparm:
sorry, but i don't see any advantages of `XOR` payloads compared to standard `AND` or `OR`? by such reasoning, we could expand to `NOT` or binary operators or arithmetic operators or...

PR #4992, opened 2022-02-18
add ord2ascii tamper script

jeffbencteux:
Changes occurrences of ORD() function to equivalent ASCII() ones.

### DIFF
--- a/tamper/ord2ascii.py
+++ b/tamper/ord2ascii.py
@@ -1,0 +1,30 @@
+#!/usr/bin/env python
+
+"""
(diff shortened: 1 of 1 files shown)


stamparm:
what's the purpose of this? i mean, you are assuming that `ORD(` is blocked, while `ASCII(` is not or?

jeffbencteux:
I used it to bypass a WAF filter on the `OR` substring. I guess the filter was used to block the `OR` SQL keyword, though it also blocked the `ORD()` function calls.

PR #5032, opened 2022-03-15
support for URLs containing credentials

the-rend:
URLs containing credentials are not supported and using them as target URLs doesn't work as expected:
```
sqlmap -u 'http://user@127.0.0.1:80/dashboard/?id=123' --batch    

[CRITICAL] host 'user@127.0.0.1' does not exist
```
```
sqlmap -u 'http://user:pass@127.0.0.1:80/dashboard/?id=123' --batch

[CRITICAL] host 'user' does not exist
```

### DIFF
--- a/lib/core/common.py
+++ b/lib/core/common.py
@@ -1721,13 +1721,13 @@ def parseTargetUrl():
 
     try:
         urlSplit = _urllib.parse.urlsplit(conf.url)
+        hostnamePort = [urlSplit.hostname, urlSplit.port]
     except ValueError as ex:
         errMsg = "invalid URL '%s' has been given ('%s'). " % (conf.url, getSafeExString(ex))
         errMsg += "Please be sure that you don't have any leftover characters (e.g. '[' or ']') "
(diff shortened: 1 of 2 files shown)


stamparm:
`conf.hostname = conf.hostname.strip("[]").replace(kb.customInjectionMark, "")` is for proper dealing with IPv6 addresses

stamparm:
also, you haven't done anything with provided username:password. what's the use of stripping those credentials?

the-rend:
`for proper dealing with IPv6 addresses`
`urlibe` supports ipv6 properly and you don't need to extract the hostname manually by splitting using `:` or striping out `[]` which results in a bug in this case.

`you haven't done anything with provided username:password`
modern browsers don't process this type of passing credentials BUT this kind of URLs are still **valid** and also you don't like adding a lot of code in one PR so I didn't implement any thing new and just patched the bug

the-rend:
so maybe I should have chosen a title like: `patch for URLs containing credentials`

Issue #4972, opened 2022-02-04
Too many open files and unable to open database file


komer83:
**Describe the bug**
Randomly during the dump process there's a too much open files and unable to open database file appears.

**To Reproduce**
1. Run 'torsocks sqlmap -u "target.com" --level=5 --risk=3 --dbs --dump -D xxx -T users --threads=10 --technique=B'
2. See error

**Expected behavior**
Under Ubuntu 21.10 with sqlmap 1.5.9#stable the error does not occur. Expected to work the same under Kali Linux and sqlmap 1.6#stable

**Screenshots**
<img width="923" alt="Screen Shot 2022-02-04 at 16 34 31" src="https://user-images.githubusercontent.com/20399230/152548035-35d8fdb8-3b6f-41a3-873e-43b79d33e7d7.png">

**Running environment:**
 - sqlmap version [e.g. 1.6#stable]
 - Installation method [e.g. apt]
 - Operating system: [e.g. Kali Linux 31-01-2022]
 - Python version [2.7.18]

**Target details:**
 - DBMS [MySQL]
 - SQLi techniques found by sqlmap [e.g. error-based and boolean-based blind]
 - WAF/IPS [none]
 - Relevant console output [in screenshot]
 - Exception traceback [if any]

**Additional context**
Altought the proccess gets terminated due to the error, running it again shows the data dumped already is stored.

stamparm:
this looks like unrelated to sqlmap. please take a look into https://github.com/nadoo/glider/issues/27 (particularly the part with `ulimit -n 102400`)

komer83:
@stamparm I have tested a retested, installed Kali on other computer. Still getting some wierd stuff here. After changing the ulimit (to any number) I'm now getting the following
<img width="612" alt="Screen Shot 2022-02-09 at 18 26 23" src="https://user-images.githubusercontent.com/20399230/153436247-6f031dfc-f524-4ee3-8db4-1d675f4780d6.png">

SQLMap works fine under Ubuntu 21.10 (no change in ulimit) with 1.5.9 but not on Kali rolling with 1.6.

stamparm:
@komer83 something is seriously wrong with your whole Kali environment. getting "buffer overflow detected", while running python program is something that can't be seen every day

komer83:
Dear @stamparm I just ran a test under Ubuntu, git cloned as instruction and ran my usual command, I still get the same error, #1.6.2.2#dev 
[...]

PR #5059, opened 2022-04-06
Fix sqlite session file descriptor leak for blind injection

CravateRouge:
Fix sqlite session file descriptor leak for blind injection, see #4972.

The cursor and connection for a thread are never closed when using `blindThread()`. It results in opening as much fd as threads without closing them when threads are finished. If the fd limit of the user is too low compared to the number of threads required, the error `unable to connect to the target URL ('Too many open files')` and `unable to open database file` will pop out and stop the DB extraction.

You can see the fd leak by monitoring the number of fd opened during the sqlmap execution with `lsof -p [sqlmap PID]`. It grows constantly. If you didn't observe this bug, check your fd limit with `ulimit -Sn` and reduce it in `/etc/security/limits.conf`.

### DIFF
--- a/lib/techniques/blind/inference.py
+++ b/lib/techniques/blind/inference.py
@@ -556,6 +556,8 @@ def bisection(payload, expression, length=None, charsetType=None, firstChar=None
                                     output += status if _ != length else " " * len(status)
 
                                     dataToStdout("\r[%s] [INFO] retrieved: %s" % (time.strftime("%X"), output))
+                        if conf.get("hashDB"):
+                            conf.hashDB.close()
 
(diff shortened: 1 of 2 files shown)


stamparm:
conf.hashDB is a global connection, and you are closing it in `blindThread`? something is wrong here

stamparm:
thank you for your report. reproduced the problem. will fix it accordingly

stamparm:
so, i had to do a more generic "patch", because also other multi-threading techniques are affected (e.g. error-based or partial union-based). thank you very much for your report/pull request :)

CravateRouge:
Glad to help :)
Thanks for your efficient maintaining on this tool!

PR #5111, opened 2022-05-29
Update README-vi-VN.md

madmandd:
Edit Vietnamese introduction for easily reading. Some technical words should not change to help readers know real meaning and can break down with google for more researching.


DemonDeath13:
it's sounds good!

stamparm:
Did a quick Google translate. Doesn't seem better than original IMHO. Also, there was no reason to change any of literal wording found in English version, like `Screenshots`-> `Overview`

madmandd:
😄 Google Translate. How funny is it!

PR #5143, opened 2022-07-17
docs: Fix a few typos

timgates42:
There are small typos in:
- thirdparty/fcrypt/fcrypt.py
- thirdparty/pydes/pyDes.py

Fixes:
- Should read `decryption` rather than `decrytpion`.
- Should read `adhered` rather than `aheared`.

Semi-automated pull request generated by
https://github.com/timgates42/meticulous/blob/master/docs/NOTE.md


stamparm:
Those are thirdparty libraries. Also, those fixes are non-code related. Thus, it is smarter to just leave them as is (to not converge from the originals)

PR #5156, opened 2022-08-02
Added the support for custom word list without _addPageTextWords

realsanjay:
Sometime what happen we knows the column name like for open source application

To fetch the specific column information, we use the custom list and that it calls that _addPageTextWords function which adds the column from website wordlist which not required 

So added another option which we can use where we have knowledge of column name for enumeration


stamparm:
sorry, but i don't see this as an improvement. not sure what's the issue with addition of `_addPageTextWords`. in such way, you could say that 50% of sqlmap's run is redundant (and you would be right), but sqlmap is made for a general use

PR #5176, opened 2022-09-15
Add URL tampering scripts for crawling

Murgeye:
While crawling, it can be helpful to rewrite URLs before visiting them. One example would be rewriting http to https or parsing a `filename` GET parameter that leads to a redirect.

This pull requests adds the config option `--url-tamper` with functionality similiar to the existing tamper script. A url-tamper script must have a function `tamper(url, **kwargs)` that returns the tampered URL.

Example URL-tamper script:
```
baseURL = example.org

# Rewrites baseURL/redirect.php?filename=$filename to baseURL/$filename
def tamper(url, **kwargs):
    if "redirect.php" in url:
        filename = url.split("filename=", 1)[1]
        new_url = "%s/%s" % (baseURL, filename)
        return new_url
    return url
```


stamparm:
sorry, but no. this doesn't look to me like something that would be of use for a general user. feel free to use it in your case

PR #5198, opened 2022-10-15
fixes #5197

### LINKED ISSUE #5197: --cookie-del option error checking seems to be incorrect
**Describe the bug**
When trying to use the `--cookie-del` option, the following error is _**always**_ returned:
image

This seems to be caused by a faulty error checking step here:
https://github.com/sqlmapproject/sqlmap/blob/70665c5d2beece0c6b71d3677fbf690f0c3695f4/lib/core/option.py#L2678-L2680

Based on the error message, it seems that only single-character `cookieDel` options should be allowed and it seems that this code is attempting to check to ensure that is the case. The condition in this `if` statement evaluates to `true` if `conf.cookieDel` exists **and** "`if len(conf.cookieDel)`" which means it's true if `conf.cookieDel` exists/is set and `len()` returns greater than `0`. Since `len()` must return greater than `0` if `conf.cookieDel` is set to anything but the empty string - which might (?) cause its own error - this if condition will always be true while trying to use the `--cookie-del` option and will always throw an error... even _**if**_ the value of the option specified is correct.

**To Reproduce**
1. Run `sqlmap -u 'someUrl' --cookie='someCookie: someVal$someOtherCookie: someOtherVal' --cookie-del='$'`
2. See error

**Expected behavior**
When `--cookie-del` is supplied with a single character, it should be accepted and used to separate the `--cookie` option value into individual cookies without throwing an error.

**Screenshots**
See "**Description**"

**Running environment A:**
 - sqlmap version: 1.6.3#pip
image

 - Installation method: Unsure, came with SANS SEC542 Ubuntu VM. Looks like maybe through snap
image

 - Operating system: Ubuntu 18.04.6 LTS
image

 - Python version unsure because I run it with the `sqlmap` command... so I suppose whatever snap uses
image

**Running environment B:**
 - sqlmap version: 1.6.10.4#dev
image

 - Installation method: `git clone`
image

 - Operating system: Windows Subsystem for linux v2 : Kali Linux 2022.1
image
image
image

 - Python version: `3.10.5` AND `2.7.18` (tested both)
image

**Target details:**
 - DBMS: mysql
 - SQLi techniques found by sqlmap: none because error on launch
 - WAF/IPS: none
 - Relevant console output: see above
 - Exception traceback: none

**Additional context**
N/A

CrazyKidJack:
Changes `--cookie-del` error checking to throw error if length of cookieDel value is not equal to **_1_**.
(It used to throw if it was not equal to **_0_** which ALWAYS returned true if the option was being used)

See issue #5197

### DIFF
--- a/lib/core/option.py
+++ b/lib/core/option.py
@@ -2675,7 +2675,7 @@ def _basicOptionValidation():
             logger.warning(warnMsg)
 
 
-    if conf.cookieDel and len(conf.cookieDel):
+    if conf.cookieDel and len(conf.cookieDel) != 1:
         errMsg = "option '--cookie-del' should contain a single character (e.g. ';')"
         raise SqlmapSyntaxException(errMsg)
 


CrazyKidJack:
You know... if `cookieDel` is _really_ supposed to be exactly ONE character... the condition should probably be `conf.cookieDel = 1` instead of `conf.cookieDel > 1`

I made it "`> 1`" in the PR because that is what was closest to the original code while still solving the problem... but "`= 1`" is probably better...

Let me know if you want me to change it before accepting

PR #5215, opened 2022-10-29
Fixing "/usr/bin/env: ‘python’: No such file or directory"

aguiar-us:
When you use "sqlmap --update" you receive this message: /usr/bin/env: ‘python’: No such file or directory

image

To fix it, only change "python" environment in sqlmap.py to "python3". You can create an environment or alias in your system called "python" to fix this issue too. The both options work.


stamparm:
sqlmap supports both python2 and python3. closing this down

PR #5253, opened 2022-12-08
added JSON waf bypass techniques

noamiscool:
I've added new WAF bypass techniques, abusing JSON syntax to bypass major WAF vendors. This was done as part of our talk in Black Hat Europe 2022 (here is a link to the research that was done in order to develop this new technique - https://claroty.com/team82/research/js-on-security-off-abusing-json-based-sql-to-bypass-waf).

### DIFF
--- a/tamper/json_waf_bypass_mysql.py
+++ b/tamper/json_waf_bypass_mysql.py
@@ -95,6 +95,10 @@ def generate_random_payload():
 
 def tamper(payload, **kwargs):
     """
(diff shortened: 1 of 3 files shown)


stamparm:
sorry, can't do this kind of self-promotion:

image

usual practice is to leave the reference (that's ok), but all greetings and stuff just can't go in

stamparm:
p.s. it should be clear from other tamper scripts how they should look like. here is a nice example with a proper reference: https://github.com/sqlmapproject/sqlmap/blob/master/tamper/dunion.py#L28

PR #5300, opened 2023-01-25
Started working on sql script.

arcad3luke:
I forked sqlmap and read through to learn how to use it and how it works, and came across a warning/exception from the editor that an import was broken, so I drilled into it and found that it was pointing to the wrong module. After running it a few times, nothing appears to have changed.


stamparm:
Thank you for report. I just pushed a minor modification to your suggestion were order of imports is preserved

arcad3luke:
Thank you!

PR #5302, opened 2023-01-30
Renamed errmsg variable

arcad3luke:
I took another gander through the file and the editor showed ~140-150 warnings for the `errmsg` variable, so I went through and renamed each one according to their origin and converted them to function strings respectively.


stamparm:
i would suggest you to either tame down editor warnings or change the editor. pushing stuff like this is not acceptable

Issue #5310, opened 2023-02-08
ReDOS Vulnerability


s0md3v:
**Describe the bug**
The regex in `lib/core/common.py` at [line 1825](https://github.com/sqlmapproject/sqlmap/blob/4357b0087d4152caba0bb04b702d9b119dfbc537/lib/core/common.py#L1825) is vulnerable to [catastrophic backtracking](https://www.regular-expressions.info/catastrophic.html). If a webpage intentionally contains a string that exploits this vulnerability, sqlmap can get stuck forever.

**To Reproduce**
The following python script can be run to see the problem in action. The first string should be matched instantaneously as expected but the second string takes about 2 minutes to match. It should be noted that the time required to match the exploit increases exponentially with each additional "X".
```
import re
pattern = '''(?i)\ASELECT(\s+TOP\s+[\d]+)?\s+\*\s+FROM\s+(([`'\"][^`'\"]+[`'\"]|[\w.]+)+)(\s|\Z)'''

print("Matching expected string")
re.search(pattern, "SELECT * FROM XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX")
print("Matched!")

print("Matching the exploit")
re.search(pattern, "SELECT * FROM XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX'")
print("Matched!")
```

**Cause & Fix**
The `[\w.]+)+)(\s|\Z)` part of the concerned regex is the culprit here, it contains adjacent repetition operators followed by a match ending condition. I have pushed a patch (#5311) that addresses this.

stamparm:
LOL

stamparm:
p.s. just for casual readers of sqlmap issues. it would be nice if author spent more than 2 minutes of reading the sqlmap code before raising this kind of LOL-ing issues. pinpointed line is part of `expandAsteriskForColumns()` which has nothing to do with parsing of web pages

PR #5311, opened 2023-02-08
ReDOS vuln patch

s0md3v:
This is a patch for the issue #5310.

I have [tested](https://regex101.com/r/GRE4CS/1) the regular expression, it works and the group number for the match is still `2` as expected. I have not however tested sqlmap itself with this change, so there's that.

### DIFF
--- a/lib/core/common.py
+++ b/lib/core/common.py
@@ -1822,7 +1822,7 @@ def expandAsteriskForColumns(expression):
     the SQL query string (expression)
     """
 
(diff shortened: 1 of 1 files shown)


stamparm:
LOL

s0md3v:
Uh something wrong? I believe it's a genuine issue.

stamparm:
https://github.com/sqlmapproject/sqlmap/issues/5310#issuecomment-1422908074

PR #5443, opened 2023-06-21
Change the dump file name

guohuan78:
Change the dump file name to `tablename-columnname-index.csv`
This method can differentiate between situations where a table exports different columns, whereas adding the sequence number to the file name does not change the file type.


kazumatzy090-sudo (commented):
[Uploading errors (1).xml…]()

PR #5501, opened 2023-08-24
Add SQLite AND boolean-based blind payload

sokaRepo:
Hello,

A contribution to add support for SQLite boolean-based using `json` function to raise exception.
At this time, there is no alternative for SQLite when standard boolean-based payload don't work and of course we prefer to avoid time-based injection.

The payload has been tested in WHERE, HAVING and GROUP BY clauses.


sokaRepo:
This PR comes along with this blog post https://sokarepo.github.io/web/2023/08/24/implement-blind-sqlite-sqlmap.html

PR #5604, opened 2024-01-12
Predict common Microsoft SQL Server users

JPSxzy8:
This is not needed when using a `defaultdict`:

```python
                if key not in kb.commonOutputs:
                    kb.commonOutputs[key] = set()
```

This was never needed (as adding the same key twice to a `set` doesn't duplicate it):

```python
                if line not in kb.commonOutputs[key]:
                    kb.commonOutputs[key].add(line)
```

### DIFF
--- a/data/txt/common-outputs.txt
+++ b/data/txt/common-outputs.txt
@@ -111,7 +111,22 @@ XDB
 
 # Microsoft SQL Server
 sa
+BUILTIN\Users
+NT SERVICE\SQLWriter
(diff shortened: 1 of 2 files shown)


stamparm:
sorry, i don't see an added value here

JPSxzy8:
Currently it is not possible to predict outputs starting-with/containing `#`, e.g. the common `##MS_SQLReplicationSigningCertificate##` users

This PR fixes that :)

PR #5649, opened 2024-03-01
Update: Chinese documents.

Nightsuki:
Add missed information from the original English version.
Revise the format and content of the previous Chinese version.


stamparm:
this doesn't look right:

image

Nightsuki:
Revised. Keep terminology: commit

Issue #5424, opened 2023-05-22
I want the result of --sql-query to be saved as a CSV file.


decuple-dream:
Hi! How are you doing? I don't know if I can make this request without a 1 dollar sponsor.
I would love to see sponsor marks on future my GitHub accounts.

As you may have noticed, the results of `-D DB -T TB -C A,B --dump` and `--sql-query="SELECT A,B FROM DB.TB" are slightly different. The latter simply saves to the output and log file.

Currently, I want to save the results to a CSV file when the -D option is not available in SQLMAP (version detection failed [I'll try to debug it!]) or when using the ORDER BY syntax(with --sql-query). 

Thank you as always. Dear Stampar👍🏼

1579091:
nice idea

System88888:
长期大量收WebShell，联系🛩TG：@System88888

anauwal:
please make this available

Aridhi-Dhia-eddine:
Hello 
I want to inderstand your prblem ; you mean that you have to use -D DB -T TB -C A,B --dump and you want to save the result in CSV File 
or you have to use  --sql-query="SELECT A,B FROM DB.TB" and you want to save the result in CSV file ?

mastercho:
>  you have to use --sql-query="SELECT A,B FROM DB.TB" and you want to save the result in CSV file ?

That

Aridhi-Dhia-eddine:
but as i know it accept only one attribute in the request is it what you want? you want to store this value in CSV file ?

Aridhi-Dhia-eddine:
hello 
just i want to know where the function "_saveToResultsFile" is called in the sqlmap project ?

AlexDevFlow:
I'd like to take this on. There's an open PR for it (#5685) that's been inactive since mid-2024, so if you're open to it I'd rather do a clean version built into the dumper.

Tracing the code, dbmsHandler.sqlQuery() hands back the values from inject.getValue() with multi-column rows already collapsed by joinValue() before the dumper sees them, and conf.dumpPath only gets created for --dump/--dump-all/--search, so a bare --sql-query has nowhere to write yet. My plan is to reuse --dump-format=csv instead of adding a flag, capture the structured rows before that join, and write them under the target's dump dir with the same safeCSValue() and --csv-del escaping a table dump uses, leaving the header off since an arbitrary SELECT has no reliable column names, unless you'd prefer something like col1,col2.

[...]

PR #5679, opened 2024-04-08
solve problem of saving result of --sql-query in CSV file

Aridhi-Dhia-eddine:
**This pull request makes the following changes:**
* Fixes issue #5424 
* i create a script (the file name is "ok.py") that return the result in a CSV file (the name of this CSV is "data_save" in the same folder with the script) when we use  (--sql-query)

**To use the script run the following command :**
-  python ok.py -u "http://testphp.vulnweb.com/artists.php?artist=1" --sql-query "SELECT aname, artist_id from acuart.artists"

image
- in the folder we have : 

image

- the content of data_save.csv is:

image


stamparm:
there won't be an inclusion of third-party scripts like this. people can always pull it from your pull-request and try to use it independetly

byt3n33dl3 (commented):
Just looking for the changes 101 (--sql-query in CSV)

PR #5690, opened 2024-04-16
Add sqlmapapi api to get admin token with user-pass.

G3G4X5X6:
# Question
I need to access the admin api of sqlmapapi remotely and it requires admin_token. But the only way to get the token is to copy it from standard output, which is inconvenient.

# Requirements

get admin_token from sqlmapapi with user-pass.


stamparm:
so, how this goes.

i'll now say "you are deliberately incorporating a security vuln", and you'll say "naah, i didn't know"

with your pull request anybody can access the admin's token. main point of "hiding" admin token in passwordless run was to prevent this kind of scenarios

stamparm:
blocked your future pull requests as i don't like this kind of "attempts". better safe than sorry

PR #5706, opened 2024-05-10
Fixed an issue with using -r to parse requests for host:ip unspliced ​​ports

Amzza0x00:
When the host in the file has a port, parseRequestFile will not splice the port to the host.


stamparm:
this is simply not true. closing this down as invalid as it would break the functionality

PR #5760, opened 2024-08-07
Create README-ckb-KU.md

SoranTabesh:
added kurdish(ckb) translation


stamparm:
Naming `[Kurdish-ckb]` doesn't look inline with other languages. Is something like Kurdish (Central)  or Kurdish (Sorani) appropriate?

SoranTabesh:
Yes
Kurdish (Central)

PR #5820, opened 2024-11-30
Fix for Google Hacking Database URL

tanaydin:


stamparm:
push the change without changing of the hash. i'll do that part

PR #5824, opened 2024-12-03
Fix for Google Hacking Database URL

tanaydin:


tanaydin:
This is for https://github.com/sqlmapproject/sqlmap/pull/5820

PR #5852, opened 2025-02-07
Added ability to add sortable tables to html dump.

tanaydin:
This PR adds to sorting links to table headers in HTML dumps if it's requested by user via command line options or settings.


tanaydin:
I'll fix this after that one merged (if it will be).
#5860 
because it's about sums.

PR #5859, opened 2025-02-19
"--fix" parameter for --smoke to update sha256 hashes

tanaydin:
This PR introduces the --fix parameter, which updates the SHA-256 hashes of modified files in the DIGEST_FILE. This parameter must be provided after --smoke.


tanaydin:
can easily done by this one liner.

git ls-files | sort | uniq | grep -Pv '^\.|sha256' | xargs sha256sum > ./data/txt/sha256sums.txt

originally from [https://github.com/sqlmapproject/sqlmap/blob/master/extra/shutils/precommit-hook.sh](https://github.com/sqlmapproject/sqlmap/blob/master/extra/shutils/precommit-hook.sh)

PR #5860, opened 2025-02-19
Adding extra parameter to smoke-test command to update sha256 sums of changed files.

tanaydin:
This PR introduces the  --update-sums parameter, which updates the SHA-256 hashes of modified files in the DIGEST_FILE. This parameter must be provided with --smoke-test.


lib/core/common.py
```diff
@@ -5599,18 +5599,34 @@ def checkSums():
 
     retVal = True
 
-    if paths.get("DIGEST_FILE"):
```
tanaydin, line 5602: getFileItem already checks for file.

PR #5874, opened 2025-03-20
Create python-package-conda.yml

WanPeirangGhoshCyber:
SQLMAP


stamparm:
too generic and there is no `environment.yml`. not sure what was the purpose of this stunt

PR #5881, opened 2025-03-26
Add luanginxmore tamper script

kennystrawnmusic:
POST requests can accept far more parameters than GET requests, so for additional evasion, it's nice to have something capable of overwhelming a WAF with millions of parameters, not just hundreds.

Tested against public bug bounty programs with great success.


stamparm:
that doctest is wrong. also, not sure why there have been any changes in the info part of the function

PR #5883, opened 2025-03-28
Fixed DEBUG variable effects in the vulnerable server code.

tanaydin:
With this change, it's checking for VULN_SERVER_DEBUG environment variable and sets to True if it's in the list of some common True values as strings. If it's set to True, it's printing some information about current process in the vulnerable server.


Dasefx19 (approved):

stamparm:
QoL for an internal auxiliary script? closing this down

PR #5893, opened 2025-04-15
Update common-files.txt and sha256sums.txt for file corrections and c…

tanaydin:
This MR removes duplicate file names in common-files.txt file with keeping first one.


stamparm:
those same files are deduplicated during the sqlmap run:

https://github.com/sqlmapproject/sqlmap/blob/master/lib/utils/brute.py#L339

PR #5910, opened 2025-06-14
Use API to check Tor connection

n-thumann:
This PR changes the way the sqlmap checks the Tor connection.
Currently, https://check.torproject.org/ is fetched and checked, whether it contains `Congratulations`.
With this PR, https://check.torproject.org/api/ip is used instead, because I think using a JSON API is much cleaner, more reliable / stable and consumes less data.

The `--check-tor` has been first implemented in https://github.com/sqlmapproject/sqlmap/commit/86b4a3562f1949398ea99dc59aaf3ae20ce0a793, before the JSON API (https://gitlab.torproject.org/tpo/network-health/metrics/tor-check/-/commit/727327d4f1f3ecd5209f5ea2e827676019c56459) has been added to check.torproject.org.

### DIFF
--- a/lib/core/option.py
+++ b/lib/core/option.py
@@ -11,6 +11,7 @@ import codecs
 import functools
 import glob
 import inspect
+import json
 import logging
 import os
 import random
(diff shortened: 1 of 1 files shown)


stamparm:
there is no check for invalid JSON:
image

PR #5930, opened 2025-07-12
Create or2logicaloroperator.py

HackingRepo:
**This tamper script bypass weak filter Filters `OR` operator via `||` operator**
# Requirement
**This script require target DBMS will be MySQL**


stamparm:
scripts having own copyright notice and/or unrelated sqlmap header will not be merged

PR #5932, opened 2025-07-14
Added support for c0-c1 as interrupt choices for quick verbosity change

tanaydin:
Quick verbosity change on keyboard interrupt choices


stamparm:
sorry, but i don't see a practical value of this for general user

tanaydin:
it was saving one enter, but it's up to you

PR #5940, opened 2025-08-13
Add support for Apache Doris database detection

lesignals:
- Add 'doris' to MYSQL_ALIASES in lib/core/settings.py
- Add DORIS fork enum in lib/core/enums.py
- Add Doris detection logic using BITMAP_UNION_COUNT() function in MySQL fingerprint
- Apache Doris will now be properly identified as "MySQL (Doris fork)" instead of showing "MySQL version not found" error

Apache Doris is highly compatible with MySQL syntax and protocol but has unique features like BITMAP functions that can be used for detection.


stamparm:
image

stamparm:
sorry, looks messed. will refactor and push smth similar

PR #5959, opened 2025-09-23
Correcting errors in the translation

arkhammknight:
Hello!
- In our language (Turkish), ‘detect’ can be used in this way "tespit etme".
- And also "exploiting" can used "bunlardan yararlanan" (benefit from these) because "istismar etmeye yarayan" means "used for abuse".
- The last one I edited "ve" that means "and".


stamparm:
image

PR #5967, opened 2025-10-18
fix logic bug in thread init

jerryjhird:
the thread finalization loop used a reversed comparison, causing the wait loop to be skipped immediately:

this change reverse the comparison so it will wait while there are active threads and elapsed time is less than the configured THREAD_FINALIZATION_TIMEOUT:


stamparm:
thx :)

PR #6021, opened 2026-02-24
implement NO_COLOR standard

jerryjhird:
implement working NO_COLOR environment variable handling
reference: https://no-color.org/


stamparm:
Closing. sqlmap already has --disable-coloring. I don’t want to add hidden behavior based on an external environment-variable convention, especially touching startup/banner/logging code for this. Not needed.

PR #6028, opened 2026-03-11
Add async mechanism for time-based blind SQLi

Saudadeeee:
Implements experimental asynchronous bisection for time-based blind SQL injection using aiohttp. Adds CLI options (--async, --async-time-based) to dramatically improve extraction speed by running binary search requests concurrently, while gracefully falling back to synchronous execution if dependencies are missing

### DIFF
--- a/lib/parse/cmdline.py
+++ b/lib/parse/cmdline.py
@@ -321,8 +321,21 @@ def cmdLineParser(argv=None):
         optimization.add_argument("--null-connection", dest="nullConnection", action="store_true",
             help="Retrieve page length without actual HTTP response body")
 
+        optimization.add_argument("--async", dest="async_opt",
+            action="store_true",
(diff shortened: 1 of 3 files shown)


stamparm:
Thank you for the effort, but sqlmap already natively supports concurrent time-based extraction via the --threads flag, which intentionally prompts a warning because parallelizing time delays inherently causes database starvation and false positives. We will not introduce a bloated third-party dependency like aiohttp to completely bypass our core request engine and reinvent a fundamentally unstable wheel.

you are even presented with a prompt when using `--threads` and time-based SQLI:

```
multi-threading is considered unsafe in time-based data retrieval. Are you sure of your choice (breaking warranty) [y/N]
```

This architectural decision is final; we consider this matter closed and will not entertain further debates or pull requests attempting to force concurrent network I/O on time-based payloads.

PR #6062, opened 2026-05-30
Fix false positives in large-response comparison fallback

ebrasha:
## What changed

This PR fixes a false-positive issue in SQLi detection when response comparison falls back for very large pages (or when `SequenceMatcher` is skipped).

Previously, in that fallback path we only compared response lengths.  
That meant two completely different large responses with similar sizes could produce a high ratio and be treated as similar, which could lead to incorrect injection signals.

The fallback now uses lightweight sampled content similarity instead of raw length-only comparison:
- Responses are normalized to bytes
- Multiple fixed windows are sampled across the response (start/middle/end)
- Byte-level similarity is calculated from sampled chunks
- Final score is combined with length ratio as a secondary signal

## Why this is better

- Reduces false positives caused by length-only matching
- Keeps fallback path fast for large responses
- Preserves existing normal comparison behavior for non-fallback paths
- Improves robustness for large JSON/CSV/binary-like payload responses

## Scope

Only `lib/request/comparison.py` is updated.  
No behavioral change intended outside the large-response/skip-seqmatcher fallback branch.

### DIFF
--- a/lib/request/comparison.py
+++ b/lib/request/comparison.py
@@ -34,6 +34,55 @@ from lib.core.settings import URI_HTTP_HEADER
 from lib.core.threads import getCurrentThreadData
 from thirdparty import six
 
+def _toBytes(value):
+    if value is None:
+        return b""
+    elif isinstance(value, six.binary_type):
+        return value
+    elif isinstance(value, six.text_type):
+        return getBytes(value, kb.pageEncoding or DEFAULT_PAGE_ENCODING, "ignore")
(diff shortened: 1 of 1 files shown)


stamparm:
Closing. This replaces a simple fallback with an arbitrary sampled-similarity heuristic. Not convinced this is generally safer, and the constants/approach look made up. Not accepting this without a real sqlmap case proving current behavior is wrong.

PR #6063, opened 2026-05-30
fix(comparison): avoid first-sample lock in matchRatio calibration

ebrasha:
## What changed

`kb.matchRatio` used to get locked on the very first valid ratio it saw.  
That was fragile — if the first response happened to be noisy (CSRF token, ad slot, timestamp widget, A/B fragment), the whole baseline for that parameter ended up biased from the start.

To fix this, I switched the calibration from single-shot to multi-sample:

- valid ratios are now collected into a small buffer instead of locking immediately
- once we hit a minimum sample count, `kb.matchRatio` gets set from the median of that buffer
- there's a hard cap on the buffer so it can't grow forever
- on context change (different page template / `textOnly` / `titles` mode), the buffer resets so old samples don't bleed into a new baseline

## Why this is better

The real problem was outliers. One bad first response could throw the whole detection logic off for that parameter. Median makes those single-sample spikes basically harmless.

In practice:
- the baseline takes a few more samples to settle, but it settles cleaner
- false positives that came from early-response noise drop a lot
- nothing else in the comparison flow gets touched — only the calibration step itself was rewritten

## Scope

Only `lib/request/comparison.py` is touched.  
The behavioral change is limited to how `kb.matchRatio` is determined.  
Detection paths, ratio bounds, and decision logic stay as-is.

### DIFF
--- a/lib/request/comparison.py
+++ b/lib/request/comparison.py
@@ -34,6 +34,67 @@ from lib.core.settings import URI_HTTP_HEADER
 from lib.core.threads import getCurrentThreadData
 from thirdparty import six
 
+MATCH_RATIO_MIN_SAMPLES = 3
+MATCH_RATIO_MAX_SAMPLES = 7
+
+def _toBytes(value):
+    if value is None:
+        return b""
+    elif isinstance(value, six.binary_type):
+        return value
+    elif isinstance(value, six.text_type):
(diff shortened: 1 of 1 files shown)


stamparm:
Closing. This is another broad heuristic change disguised as a small calibration fix. It changes core comparison behavior and even adds unrelated similarity logic. Not accepting fan-code without a concrete reproducible sqlmap problem.

PR #6064, opened 2026-05-30
fix(comparison): make decision tolerance adaptive to observed page jitter

ebrasha:
## What changed

The final decision in `_comparison()` used a hard-coded `DIFF_TOLERANCE = 0.05`.  
On dynamic targets, normal page noise (rotating banners, timestamps, ad slots, slight markup drift) easily crosses that 5% line and gets read as a real True/False signal.

To fix this, I made the tolerance adaptive:

- when the median-based `kb.matchRatio` gets locked, I also compute jitter using MAD (Median Absolute Deviation) over the same calibration buffer
- jitter goes into `kb.matchRatioJitter`
- the final decision now uses `tolerance = max(DIFF_TOLERANCE, JITTER_TOLERANCE_MULTIPLIER * jitter)`
- on context change, both buffer and jitter get reset so old noise doesn't bleed into a new baseline

The static `DIFF_TOLERANCE` still acts as a hard floor — pages that are actually stable behave exactly like before.

## Why this is better

5% was a reasonable default 10 years ago, but on modern dynamic pages it's just too tight. We were paying the cost in false positives.

- using MAD keeps the noise estimate robust against single bad samples (same reason we use median for the baseline itself)
- multiplier of 3.0 with MAD gives roughly ~2σ-equivalent for normally distributed noise, which is a sane signal-detection threshold
- stable targets see no change because the floor still kicks in
- noisy targets get a wider band that actually matches their real noise level, so we stop labeling random page drift as a successful boolean payload

## Scope

`lib/request/comparison.py`

### DIFF
--- a/lib/request/comparison.py
+++ b/lib/request/comparison.py
@@ -34,6 +34,85 @@ from lib.core.settings import URI_HTTP_HEADER
 from lib.core.threads import getCurrentThreadData
 from thirdparty import six
 
+MATCH_RATIO_MIN_SAMPLES = 3
+MATCH_RATIO_MAX_SAMPLES = 7
+
+# Multiplier applied to the observed jitter (MAD) when computing the
+# adaptive decision tolerance. Roughly mimics ~2 sigma for a normal
+# distribution while staying robust against outliers.
+JITTER_TOLERANCE_MULTIPLIER = 3.0
+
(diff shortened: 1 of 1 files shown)


stamparm:
this looks like "fan proposition". closing this down without too much explanation

PR #6069, opened 2026-06-05
feat(db2): add light time-based payloads

Giardi77:
Adds lower-impact IBM DB2 time-based blind payloads that use a two-way SYSIBM.SYSTABLES Cartesian product before the existing three-way heavy-query payloads.

This preserves the existing heavier DB2 payloads as fallbacks while giving sqlmap a cheaper first timing probe for DB2 targets where the current payload can be too disruptive.

Also extends the payload loader doctest to assert that the DB2 light payloads are loaded and do not use the T3 alias.

Verification:
- python3 -c "from xml.etree import ElementTree as et; et.parse('data/xml/payloads/time_blind.xml'); print('xml ok')"
- python3 sqlmap.py --smoke


stamparm:
nope. DB2 is already over-presented in time-based payloads

PR #6071, opened 2026-06-06
Fix Oracle current_db query to return actual database name

zakaria-zoulati:
Previously used SELECT USER FROM DUAL which returns the username,
not the database name. Replaced with ORA_DATABASE_NAME which returns
the actual DB name for any user on Oracle 12c+. USER kept as query2
fallback for older Oracle versions.

Tested on Oracle 21c XE via Docker (gvenzl/oracle-xe:21-slim)"

image


stamparm:
Breaks sqlmap's Oracle enumeration model and hands -D users bad info

PR #6088, opened 2026-07-28
feat(tamper): add space2tab tamper script for WAF bypass via tab substitution

Char0n1507:
## Summary
Adds a new tamper script `space2tab.py` that replaces space characters
with tab characters (\t / %09).

## Use Case
Some WAFs block %20 (URL-encoded space) but allow %09 (URL-encoded tab),
which is valid whitespace in MySQL, MSSQL, PostgreSQL, Oracle and SQLite.

## Behaviour
- Spaces outside string literals are replaced with \t
- Spaces inside single/double quoted strings are preserved
- None/empty payloads are returned as-is

## Tests
tamper('SELECT id FROM users')              -> 'SELECT\tid\tFROM\tusers'
tamper('1 AND 1=1')                         -> '1\tAND\t1=1'
tamper("WHERE name='John Doe'")             -> "WHERE\tname='John Doe'"
tamper(None)                                -> None

### DIFF
--- a/tamper/space2tab.py
+++ b/tamper/space2tab.py
@@ -1,0 +1,63 @@
+#!/usr/bin/env python
+
+"""
+Copyright (c) 2006-2026 sqlmap developers (https://sqlmap.org)
+See the file 'LICENSE' for copying permission
+"""
+
+from lib.core.compat import xrange
+from lib.core.enums import PRIORITY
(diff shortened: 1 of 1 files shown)


stamparm:
tons of similar tamper scripts are already inside

PR #6119, opened 2026-09-14
Fix DBMS name/version reporting fallback and resume detection

tanaydin:
## Summary
- `Format.getDbms()` / `Backend.getDbms()` now fall back to `conf.dbms` when `kb.dbms` isn't set yet, so a formatted DBMS+version string no longer drops the DBMS name or prints "None" before fingerprinting completes.
- `_resumeDBMS()` in `lib/core/target.py` now also propagates the DBMS/version into `Backend` when resuming a session where the DBMS wasn't previously stored, and preserves the previously detected version when the resumed value is unknown.
- Minor simplification cleanups in `lib/core/option.py` (non-SQL technique switch validation, tamper priority check, `kb.chars`/`kb.multibit` initialization).
- Added unit tests in `tests/test_misc.py` covering the `Format.getDbms()` fallback behavior.

## Test plan
- [x] `python tests/test_misc.py` (new `TestFormatGetDbms` cases pass)


stamparm:
that is not how this is done. changing 4 files for sqlmap, without even opening an issue or asking whether you could do it will always result with close down

stamparm:
you really think that i could merge something like `Minor simplification cleanups in lib/core/option.py (non-SQL technique switch validation, tamper priority check, kb.chars/kb.multibit initialization).` without any issue where you explain "your problem" first? no way

tanaydin:
Ok, I'll try to provide a working test for this, it's hard to reproduce, "sometimes" I'm hitting that bug.

stamparm:
man, you open an issue describing your "problem", then you leave me to try to fix it. if i get stuck (i doubt it), you can do your pull request. that's the way this story goes. this is wrong repo for gaining merge points

tanaydin:
not looking for gain any points. will track correct way. thank you for explanation.

